PR #226, opened 2026-03-25
feat(introspection): implement OAuth 2.0 Token Introspection (RFC 7662)

### LINKED ISSUE #16: Introspection

jamescrowley321:
## Summary

Implement OAuth 2.0 Token Introspection per RFC 7662:

- **`TokenIntrospectionRequest/Response`** models extending `BaseRequest/BaseResponse`
- **`introspect_token()`** sync + async with DI and retry support
- Client authentication via Basic auth (confidential) or body params (public)
- Response includes `active` bool and optional metadata claims

## Test Plan

- [x] 7 unit tests: active/inactive tokens, error, token_type_hint, public client, model inheritance
- [x] 2 integration tests: discovery endpoint check, request model from discovery
- [x] Usage example
- [x] 369 total tests pass, 90.10% coverage

Automated PR from ralph loop task T37. Closes #16


jamescrowley321:
## Adversarial Code Review — PR #226

### Layer 1: Blind Hunter

1. **`parse_introspection_response` not in `response_processors.__all__`** — `parse_introspection_response` is defined in `core/response_processors.py` but not added to the module's `__all__` list (lines 353-360). Every other parser (`parse_auth_code_token_response`, `parse_jwks_response`, `parse_token_response`, `parse_userinfo_response`) is in `__all__`. This is inconsistent and means the function is technically a private implementation detail despite being imported by `core/introspection_logic.py`.

2. **`handle_introspection_error` not in `error_handlers.__all__`** — Same pattern: `handle_introspection_error` is defined in `core/error_handlers.py` but not added to `__all__` (lines 214-220). Every other handler (`handle_auth_code_token_error`, `handle_discovery_error`, `handle_jwks_error`, `handle_token_error`, `handle_userinfo_error`) is listed. Inconsistent with existing conventions.

3. **No `response.json()` failure handling in `parse_introspection_response`** — If the introspection endpoint returns HTTP 200 with a non-JSON body (malformed response, HTML error page behind a reverse proxy, etc.), `response.json()` at `response_processors.py:302` will raise a `JSONDecodeError`. This exception propagates up to `process_introspection_response` in `introspection_logic.py:59-66`, where the catch-all `except Exception` calls `handle_introspection_error`. So it is caught, but the error message will say "Unexpected error during token introspection" with a raw JSONDecodeError string rather than a clear "invalid JSON" message. Not a crash, but a poor diagnostic experience. Other parsers in this codebase (e.g., discovery) explicitly catch `ValueError` from `.json()` and raise a domain-specific exception with a clear message.

4. **`response.content` in error messages is `bytes`, not `str`** — In `parse_introspection_response` (line 307), `response.content` is typed as `bytes`. The f-string interpolation will produce something like `b'Unauthorized'` in the error message. This is consistent with every other parser in `response_processors.py` (they all do the same thing), so it is a pre-existing pattern, but it is still a poor UX pattern that this PR perpetuates.

5. **No async unit tests** — The PR adds async implementation (`aio/introspection.py`) but only sync unit tests (`tests/unit/test_introspection.py`). Every other feature in this codebase has both sync and async unit test files (e.g., `test_sync_userinfo.py` / `test_aio_userinfo.py`, `test_sync_token_client.py` / `test_aio_token_client.py`). The async path has a subtle difference (no `response.close()` call) that is untested.

6. **No network error unit test** — The existing unit tests cover happy paths (active/inactive token), HTTP error (401), token type hint, and public client. But there is no test for network-level failures (e.g., `httpx.ConnectError`). Every comparable module's test suite includes a `test_*_network_error` test (see `test_sync_userinfo.py:95`, `test_aio_userinfo.py:88`, `test_sync_token_client.py:86`, `test_aio_token_client.py:67`). The `handle_introspection_error` code path for `httpx.RequestError` is untested.

7. **No `FailedResponseAccessError` guard test** — `TokenIntrospectionResponse` declares `_guarded_fields = frozenset({"claims"})`, meaning accessing `.claims` on a failed response should raise `FailedResponseAccessError`. This guard behavior is not tested. Comparable test suites (e.g., `test_aio_userinfo.py:82-83`) explicitly test this with `pytest.raises(FailedResponseAccessError)`.

8. **`client_id` logged at DEBUG level without redaction** — In `introspection_logic.py:20`, the client_id is logged via `logger.debug(f"Client ID: {request.client_id}")`. While client_id is not a secret per se, in combination with the endpoint URL it can be used for targeted attacks. The URL is redacted via `redact_url()` but client_id is logged in plaintext. This is minor but inconsistent with the security posture of redacting the URL.

9. **Double error wrapping in `process_introspection_response`** — `introspection_logic.py:59-66` wraps `parse_introspection_response` in a try/except that calls `handle_introspection_error`. But the callers (`sync/introspection.py:58`, `aio/introspection.py:56`) also wrap the entire flow including `process_introspection_response` in their own try/except that calls `handle_introspection_error`. If `parse_introspection_response` raises, the exception is caught and handled at the `process_introspection_response` level, and the caller's try/except never fires. This is not a bug (both paths produce the same result), but the inner try/except is redundant defensive code that obscures flow.

10. **`token` field accepts empty string without validation** — `TokenIntrospectionRequest.token` is typed as `str` with no validation. An empty string `""` is a valid `str` and would be sent to the introspection endpoint as `token=` in the POST body. RFC 7662 Section 2.1 says the token parameter is REQUIRED, but no client-side validation prevents sending an empty value. Same applies to `client_id` and `address`.


### Layer 2: Edge Case Hunter

1. **`response.json()` returns non-dict** — `parse_introspection_response` at `response_processors.py:299-303` calls `response.json()` and assigns the result directly to `claims`. If the server returns a JSON array (`[]`) or scalar (`true`, `42`), `claims` will be a `list` or `bool`/`int`, not a `dict`. Downstream code in `process_introspection_response` at `introspection_logic.py:61` calls `result.claims.get("active", False)` which will raise `AttributeError` on non-dict types. The catch-all in `introspection_logic.py:65` converts this to an error response, but the error message will be confusing ("Unexpected error... 'list' object has no attribute 'get'").

2. **`claims` is empty dict `{}`** — If the server returns `{}` (no `active` field), `process_introspection_response` at `introspection_logic.py:61` evaluates `result.claims` as truthy (empty dict is falsy in Python, actually). Wait — correction: empty dict `{}` is falsy. So the condition `if result.is_successful and result.claims:` at line 61 would be `False`, and the `active` status would never be logged. The response would still be returned with `is_successful=True` and `claims={}`. The caller would then need to check `claims.get("active", False)` themselves. This is not a bug per se, but it means a response missing the RFC-required `active` field silently passes as successful.

3. **Server returns HTTP 200 with `Content-Type: text/html`** — Some authorization servers behind reverse proxies may return an HTML error page with 200 status. `response.json()` will raise `JSONDecodeError`. As noted above, this is caught by the double-layer exception handling, but produces a confusing error message.

4. **Retry behavior on 401/403** — The `_introspect_token` function is wrapped with `@retry_with_backoff()`. If the introspection endpoint returns 401 (bad credentials) or 403 (forbidden), the retry logic's `should_retry_response` determines whether to retry. Since 401/403 are not 429 or 5xx, they should not be retried. This is correct behavior inherited from the shared retry infrastructure, but worth noting that it is implicitly correct rather than explicitly tested for introspection.

5. **`http_client.client` is `None`** — If the caller passes an `HTTPClient`/`AsyncHTTPClient` instance where `.client` is `None` (e.g., client was closed or not initialized), `sync/introspection.py:51` or `aio/introspection.py:51` will set `client = None`, and the subsequent `client.post(...)` will raise `AttributeError: 'NoneType' object has no attribute 'post'`. This is caught by the outer try/except, but the error message will be misleading. This is a pre-existing pattern across all modules, not specific to this PR.

6. **Large response body in error message** — When `parse_introspection_response` handles a non-success response, the error message at `response_processors.py:306-308` includes `response.content`, which could be arbitrarily large (e.g., a verbose HTML error page). This could cause log bloat or memory issues in extreme cases. This is a pre-existing pattern across all parsers.


### Triage Summary

#### MUST FIX

*None.* The implementation is structurally sound and follows existing patterns. No security vulnerabilities, no data loss risks, no crashes in expected paths.

#### SHOULD FIX

**SF-1. Add `parse_introspection_response` to `response_processors.__all__` and `handle_introspection_error` to `error_handlers.__all__`** (Blind Hunter #1, #2)
- Files: `src/py_identity_model/core/response_processors.py:353-360`, `src/py_identity_model/core/error_handlers.py:214-220`
- Every other parser and handler is in `__all__`. This is a consistency gap that should be closed before merge.

**SF-2. Add async unit tests for `aio/introspection.py`** (Blind Hunter #5)
- Every other feature has `test_aio_*.py` alongside `test_sync_*.py`. The async introspection path (which notably omits `response.close()`) is completely untested.

**SF-3. Add network error and `FailedResponseAccessError` guard unit tests** (Blind Hunter #6, #7)
- Missing tests for `httpx.ConnectError` side effect and for accessing `.claims` on a failed response. Both patterns are tested in every comparable module.

#### DEFER

**D-1. `response.json()` failure produces a confusing error message** (Blind Hunter #3, Edge Case #1, #3)
- Pre-existing pattern: no parser in `response_processors.py` guards against non-JSON 200 responses. The double-layer exception handling prevents crashes but yields unhelpful messages. Not introduced by this PR and not worth fixing in isolation.

**D-2. `response.content` as `bytes` in error messages** (Blind Hunter #4, Edge Case #6)
- Pre-existing pattern across all parsers. Not introduced by this PR.

**D-3. No client-side validation of `token`, `client_id`, or `address`** (Blind Hunter #10)
- Pre-existing pattern: `BaseRequest.address` is an unvalidated `str` across the entire codebase. Not a regression.

**D-4. Empty dict `{}` or missing `active` field silently passes** (Edge Case #2)
- RFC 7662 requires `active` in the response, but validating server compliance is not this library's responsibility at this layer. The caller is expected to check `claims["active"]`.

**D-5. Double error wrapping in `process_introspection_response`** (Blind Hunter #9)
- Pre-existing pattern used by other `*_logic.py` modules. Redundant but harmless.

**D-6. `client_id` logged without redaction** (Blind Hunter #8)
- Minor. `client_id` is a public identifier per OAuth 2.0 spec, not a secret. Consistent with other modules that log similar identifiers.

Issue #242, opened 2026-03-27
OpenID Connect RP Certification — Tracking Issue


jamescrowley321:
## Objective

Achieve **OpenID Connect Certified** status for **py-identity-model as a Relying Party library** through the OpenID Foundation's self-certification program, following the same deployment model that Roland Hedberg used for `pyoidc` / `oidcrp` and Dominick Baier used for `IdentityModel.OidcClient`.

**Certification target:** the py-identity-model library itself, listed as:
`py-identity-model <version> — James Crowley — Python — Apache 2.0 — [profiles certified]`

**Scaffolding (not certified):** `conformance/app.py` is a disposable FastAPI test harness that drives the library through the OIDF conformance suite, exactly analogous to Duende's [`clients/ConformanceTests/`](https://github.com/DuendeArchive/IdentityModel.OidcClient/tree/main/clients/ConformanceTests) C# console runner. The harness is not shipped, not versioned, and never becomes a product — it's a test vehicle.

---

## Plan revision history

- **2026-04-11 (revised)** — rewritten after adversarial re-review of #322/#324/#325 and public research into: (a) the pyoidc / oidcrp / Duende cert precedents, (b) certification.openid.net REST API availability, (c) CLI and native-app certification constraints, (d) OIDF fee structure and OSS waiver policy. See research summary at the bottom of this issue.
- **Original plan (now superseded)** assumed we needed to build a FastAPI web-app deployment and a CLI deployment as two separate $700 cert targets with Playwright-driven UI automation. That framing was wrong: libraries are certifiable directly, certification.openid.net exposes the same REST API as the self-hosted suite, and only one $700 fee is needed (or $0 via OSS waiver).

---

## Deployment model — what we're actually certifying

**We are certifying the library, not a product.**

| Role | Artifact | Shipped? | Certified? |
|---|---|---|---|
| **Certified deployment** | `py-identity-model` the Python package | Yes, PyPI | **Yes** — the cert listing names this |
| **Disposable test harness** | `conformance/app.py` + `run_tests.py` + docker-compose.yml | No — repo only | No — scaffolding |
| **Downstream product** | `py-identity-model-cli` (new CLI) | Yes, future | No — inherits credibility from library cert |
| **Downstream product** | `fastapi-identity-model` (new FastAPI middleware) | Yes, future | No — inherits credibility from library cert |
| **Regression shield** | `.github/workflows/conformance.yml` against local Authelia suite | — | — |

[...]

PR #321, opened 2026-04-10
feat(conformance): add OIDF RP conformance test harness (#242)

jamescrowley321:
## Summary

- Add thin FastAPI RP harness (`conformance/app.py`) that exercises py-identity-model's sync public API (discovery, token exchange, ID token validation, UserInfo) against the OpenID Foundation conformance suite
- Add Docker Compose stack (`conformance/docker-compose.yml`) with MongoDB, Java conformance server, nginx TLS termination, and the RP container — all wired for `devmode=true` local testing
- Add test runner (`conformance/run_tests.py`) that automates plan creation, test execution, polling, and result collection via the suite REST API
- Add Makefile targets: `conformance-build`, `conformance-up`, `conformance-down`

## Conformance Results

N/A — harness infrastructure only. Actual conformance runs are tracked in subsequent tasks (C2–C4).

## Test Results

- Unit tests: 840 passed
- Coverage: 95.28%
- `make lint`: all hooks passed (ruff lint, ruff format, pyrefly typecheck)


conformance/nginx/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM nginx:1.27.3
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends openssl \
+    && openssl req -x509 -nodes -days 3650 -newkey rsa:2048 \
+        -keyout /etc/ssl/private/nginx-selfsigned.key \
+        -out /etc/ssl/certs/nginx-selfsigned.crt \
+        -subj "/CN=localhost.emobix.co.uk" \
```
jamescrowley321, line 8: What is this subject? We already have scripts for self signed certs in the repo.

conformance/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /app
+
+# Install harness dependencies
+COPY conformance/requirements.txt requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Install py-identity-model from source
+COPY pyproject.toml README.md ./
+COPY src/ src/
+RUN pip install --no-cache-dir .
```
jamescrowley321, line 12: Duplicating effort with dockerfiles and we use uv.

jamescrowley321:
## Fresh-Context Adversarial Review — #321 feat(conformance): add OIDF RP conformance test harness

Reviewed under fresh context with no prior knowledge of the PR author's intent. Five personas applied sequentially.

---

### Blind Hunter

**MUST FIX**

- [`conformance/app.py:251`] `assert disco.authorization_endpoint is not None` — a bare `assert` in production-path request handling. Python strips `assert` statements when run with `-O`/`-OO` (e.g. in some Docker base images with `PYTHONOPTIMIZE=1`). If the conformance suite returns a discovery doc without `authorization_endpoint`, this silently becomes `None`, `build_authorization_url` receives `None`, and the behavior is undefined rather than returning a clean 502. Same issue at lines 345 and 370 (`assert disco.token_endpoint is not None`, `assert token_response.token is not None`). Replace all three with explicit guard+return.

- [`conformance/app.py:379`] `clear_jwks_cache()` is called unconditionally on every callback, for every test. The cache is a module-global dict (`_jwks_cache` in `sync/token_validation.py`). In concurrent test runs (the runner is sequential, but the harness is multi-process-safe only if not shared), clearing before every validate defeats the key-rotation test design: `rp-key-rotation-op-sign-key` expects the harness to *attempt* validation with the cached (stale) key first, *then* re-fetch on failure. Clearing upfront means the test always fetches fresh keys and never exercises the rotation-retry path. This is a conformance shortcut masquerading as a correctness fix.

**SHOULD FIX**

- [`conformance/app.py:181`] `sessions: dict[str, AuthSession] = {}` — global in-memory session store with no eviction, no TTL, no size cap. A misbehaving conformance suite test that never completes a callback leaks one `AuthSession` per triggered authorize call, forever, for the lifetime of the process. Not dangerous in a sandboxed test run, but worth noting.

- [`conformance/app.py:291`] `return JSONResponse(status_code=502, ...)` inside a function whose return type annotation is `HTMLResponse | JSONResponse` — that's fine. But the caller `callback_get` has return type `HTMLResponse | JSONResponse` too, and the `authorize` endpoint also returns `RedirectResponse`. FastAPI will coerce these correctly, but the type annotation on `authorize` says `-> RedirectResponse` while two of its branches return `JSONResponse`. This will cause type errors with strict checkers and is misleading to callers.

- [`conformance/run_tests.py:755`] `client_secret` passed as a plain query parameter to the RP `/authorize` endpoint. It travels in the URL (and thus in proxy/nginx access logs) on every test invocation. It's a local dev-only secret so the blast radius is small, but the pattern is wrong regardless.

**NITPICK**

- [`conformance/app.py:186-194`] Singleton HTTP client via `global` + module-level `None` initialization — suppressed by the `PLW0603` ruff ignore already added. The pattern works, but the singleton is not cleaned up on process exit and its `verify=False` setting cannot be changed after first initialization.

---

### Edge Case Hunter

| `conformance/app.py:297` | `cb_response.state` is `None` and is not in `sessions` — but `sessions.pop(None)` would be a `KeyError` if the `None not in sessions` guard is reached but the subsequent `sessions.pop(state_value)` is called | Already guarded by the `if state_value is None or state_value not in sessions` check — **no bug** | — |
| `conformance/app.py:372–415` | Token response has no `id_token` field — `id_token_jwt` is `None`, the `if id_token_jwt:` block is skipped entirely, `claims` stays `{}`, nonce is **never validated** | `if not id_token_jwt: return HTMLResponse(content="...", status_code=400)` | **[WRONG]** flow succeeds without nonce or ID token validation; conformance test for `rp-nonce-invalid` would incorrectly pass |
| `conformance/app.py:403–415` | Token nonce claim absent (`claims.get("nonce")` returns `None`) while `session.nonce` is a 32-byte random string — comparison `None != session.nonce` correctly catches this | No fix needed — logic is correct for absent nonce | — |
| `conformance/app.py:467–476` | `form_data` values from `request.form()` may be `UploadFile` objects, not strings — `dict(form_data)` can include file handles; `urlencode` then serializes `<UploadFile ...>` as the value | `params = urlencode({k: str(v) for k, v in form_data.items() if not hasattr(v, 'read')})` | **[WRONG]** malformed callback URL passed to `_handle_callback`; parse will succeed but with garbage parameter values |
| `conformance/run_tests.py:892` | `module_info.get("id", module_info.get("name", ""))` — if both `id` and `name` are absent, `module_id` is `""`. Subsequent `suite.start_test("")` calls `POST /api/runner/` (trailing slash, empty path segment) — suite returns 404, caught and reported as FAILED | `if not module_id: return TestResult(test_name=test_name, test_id="", status="FAILED", detail="Suite returned no module ID")` | **[WRONG]** misleading FAILED result with no actionable detail |
| `conformance/run_tests.py:1003` | `plan_response.get("modules", [])` — if the suite returns modules as a non-list type (e.g. `null`), iteration fails with `TypeError` | `modules = plan_response.get("modules") or []` | **[CRASH]** `TypeError: 'NoneType' object is not iterable` |

**Summary**
- Unhandled paths found: 4 real
- Critical (crash/data loss): 1 (`run_tests.py:1003`)
- Non-critical (wrong result/degraded): 3 (`app.py:372`, `app.py:467`, `run_tests.py:892`)

---

### Acceptance Auditor

The issue (#242) Phase 2 acceptance criteria, extracted:

- **AC-1**: Build thin FastAPI test harness in `conformance/` using py-identity-model's discovery, JWKS, token validation, and UserInfo
- **AC-2**: Local conformance suite Docker Compose (OIDF Java server + MongoDB + nginx TLS)
- **AC-3**: Automated test runner via suite REST API (`/api/plan`, `/api/runner`, `/api/info`)
- **AC-4**: Must enable running Basic RP certification plan (`oidcc-client-basic-certification-test-plan`)
- **AC-5**: Must enable running Config RP certification plan (`oidcc-client-config-certification-test-plan`)
- **AC-6**: Must use py-identity-model's **public API only** (no internals access)

**PASS**

- [AC-1] FastAPI RP harness — implemented at `conformance/app.py:149`, uses `get_discovery_document`, `request_authorization_code_token`, `validate_token`, `get_userinfo`, `parse_authorize_callback_response`, `validate_authorize_callback_state`, `build_authorization_url`, `generate_pkce_pair`, all from public `py_identity_model` namespace
- [AC-2] Docker Compose — `conformance/docker-compose.yml` with MongoDB 6.0.13, `authelia/openid-connect-conformance-suite:latest`, nginx TLS at 8443
- [AC-3] Test runner — `conformance/run_tests.py` implements plan creation, module creation, start, polling via `ConformanceSuiteClient` class
- [AC-4] Basic RP plan — `conformance/configs/basic-rp.json` references `oidcc-client-basic-certification-test-plan`; runner drives authorize flow per module
- [AC-5] Config RP plan — `conformance/configs/config-rp.json` references `oidcc-client-config-certification-test-plan`
- [AC-6] Public API — all imports are from `py_identity_model` top-level or `py_identity_model.exceptions`; no `from py_identity_model.core.*` or `from py_identity_model.sync.*` imports

**PARTIAL**

- [AC-1 / nonce] Nonce validation is implemented **only when `id_token` is present** (lines 372–415). When `id_token_jwt` is `None`, nonce is skipped entirely and flow is marked successful. The OIDC spec requires `id_token` for code flow; the conformance suite will always return one. This is latent rather than observable in practice, but the code path exists and would pass a test it should fail.

**FAIL**

- None that block the stated scope (harness infrastructure, not actual conformance runs).

**SCOPE CREEP**

- `src/py_identity_model/__init__.py` and `src/py_identity_model/sync/__init__.py` — `clear_jwks_cache` is added to the public library API to serve the harness. This is a library change motivated by a harness concern. The harness does not *need* `clear_jwks_cache` in the public API — it existed in `sync/token_validation.py` but was not exported. Exposing it pollutes the public surface and sets a precedent for test-driven API additions. This should be documented as a deliberate design decision, or the harness should use an alternative pattern.

**Architecture Violations**

- None that violate documented mandatory guidelines.

**Summary**
- Total ACs: 6
- Pass: 5 | Fail: 0 | Partial: 1

---

### Sentinel (Security)

**BLOCK**

None. The harness is explicitly a local-only dev/test tool and is not part of the library's production surface.

**WARN**

- [LIKELY] `conformance/app.py:159` — `RP_BASE_URL` defaults to `http://localhost:8888` (plain HTTP). The harness accepts `client_secret` as a query parameter to `/authorize` (line 212). If the RP is started with a non-localhost `RP_BASE_URL` (e.g. on a shared CI host), the secret travels in plain HTTP query strings and appears in access logs. The Docker Compose does not bind-restrict the RP port (`8888:8888` without `127.0.0.1:`), so the RP is reachable from other containers and from the host network.
  Mitigation: Bind RP port to `127.0.0.1:8888:8888` in `docker-compose.yml`, and pass client credentials via POST body or request header rather than URL query params.

- [UNLIKELY] `conformance/app.py:379` — `clear_jwks_cache()` is a process-global operation. If the harness were ever run in a context shared with a production validate-token path (it won't be, but the API is now public), clearing the cache mid-validation would cause a cache miss and a potentially unexpected re-fetch.
  Mitigation: document that `clear_jwks_cache` is intended for test use only.

**INFO**

- `conformance/nginx/Dockerfile:8` — Self-signed cert generated at image build time (not at container start), so the cert is baked into the image layer. For local dev this is fine. For any shared image registry, the private key would be in the image. Acceptable given the explicit dev-only scope.
- `conformance/docker-compose.yml:562` — `SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_GITLAB_CLIENT_ID=dummy` / `CLIENT_SECRET=dummy` — dummy credentials for devmode; no real secret.
- `conformance/app.py:193` — `HTTPClient(verify=False)` — TLS verification disabled. Required for the self-signed cert setup; acceptable for the local conformance harness. Suppressed by ruff `S501` ignore already present.

**Summary**
- BLOCK: 0 | WARN: 2 | INFO: 3
- Overall: PASS (with WARN follow-up)

---

### Viper (Red Team)

**Attack Surface**

- `GET /authorize` — unauthenticated endpoint accepting arbitrary `issuer` and `client_secret` query params
- `GET /callback` + `POST /callback` — unauthenticated, accepts arbitrary URL data
- `GET /results/{test_id}` — unauthenticated read of stored test results
- Docker Compose: RP port 8888 bound to all interfaces (not 127.0.0.1)
- Docker Compose: MongoDB port not exposed externally — correct
- `authelia/openid-connect-conformance-suite:latest` — mutable tag; supply chain risk if image is replaced

#### MEDIUM — SSRF via user-controlled `issuer` parameter

- **Location**: `conformance/app.py:225`
- **Attack scenario**:
  1. Attacker has network access to the host running the conformance harness (possible given `0.0.0.0:8888` binding)
  2. Attacker sends `GET /authorize?issuer=http://169.254.169.254/latest/meta-data/&client_id=x`
  3. `get_discovery_document` makes an outbound HTTP GET to the attacker-controlled URL; `verify=False` is set so TLS errors don't block the request
  4. The response body is parsed as a discovery document; partial metadata leaks in error messages (e.g. `discovery_failed: <response body excerpt>`)
  5. Attacker iterates to probe internal network services reachable from the RP container
- **Prerequisites**: Network access to port 8888 of the host running the harness
- **CVSS v3.1**: 5.8 (AV:N/AC:H/PR:N/UI:N/S:C/C:L/I:N/A:N) — elevated by scope change (container to host network)
- **Remediation**: Validate `issuer` against an allowlist of expected conformance suite issuer prefixes (e.g. `https://localhost.emobix.co.uk:8443/`), or bind the RP to `127.0.0.1` only. The Docker Compose fix is: `ports: ["127.0.0.1:8888:8888"]`.

#### LOW — test result disclosure via predictable `test_id`

- **Location**: `conformance/app.py:479`
- **Attack scenario**: The `test_id` is the conformance suite's module ID (a UUID or short string). `GET /results/{test_id}` returns the full session result including `access_token`. An attacker who can guess or enumerate module IDs (from the suite's public `/api/info` endpoint, which is unauthenticated in devmode) can retrieve access tokens from completed flows.
- **Prerequisites**: Network access to port 8888; knowledge that the conformance suite is running
- **CVSS v3.1**: 3.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N)
- **Remediation**: Either remove access_token from the stored result, or scope `/results` to localhost only.

#### LOW — `authelia/openid-connect-conformance-suite:latest` mutable image tag

- **Location**: `conformance/docker-compose.yml:557`
- **Attack scenario**: If the Authelia registry entry for `latest` is compromised or silently updated, `docker compose pull` on a developer's machine pulls a malicious image that acts as a rogue OP and issues tokens the RP will validate.
- **Prerequisites**: Attacker controls or compromises the image registry
- **CVSS v3.1**: 4.2 (AV:N/AC:H/PR:N/UI:R/S:U/C:L/I:L/A:N)
- **Remediation**: Pin to a specific digest or version tag.

**Summary**
- Attack surface elements: 5
- Overall: PASS (all findings are acceptable for an explicitly local dev tool — WARN on SSRF if any CI environment exposes port 8888)

---

### Triage Summary

| Priority | Finding | Location |
|----------|---------|----------|
| **P1** | `clear_jwks_cache()` on every callback defeats key-rotation conformance test | `app.py:379` |
| **P1** | `assert` in request-handling path — stripped under `-O` optimization | `app.py:251, 345, 370` |
| **P2** | `id_token` absent → nonce skipped entirely, flow succeeds | `app.py:372–415` |
| **P2** | `form_data` dict values may be `UploadFile`, not `str` | `app.py:474` |
| **P2** | `modules` from plan response may be `None` → `TypeError` | `run_tests.py:1003` |
| **P2** | SSRF via user-controlled `issuer` + `0.0.0.0:8888` binding | `app.py:225`, `docker-compose.yml` |
| **P2** | `clear_jwks_cache` exposed as public library API to serve harness | `src/py_identity_model/__init__.py` |

**P0**: 0
**P1**: 2
**P2**: 5

**Overall gate**: ⚠️ MERGE WITH FOLLOWUP

The harness infrastructure is sound and the public API is used correctly throughout. The two P1 issues should be addressed before running actual conformance tests (the `assert` issue is latent; the `clear_jwks_cache` issue will cause the key-rotation tests to produce false results). The P2 items are follow-up worthy but do not block the harness from being built and iterated on in subsequent tasks (C2–C4 as referenced in the PR).

jamescrowley321:
Superseded by #322 — the `feat/basic-rp-conformance` branch chains from this branch and contains all 4 commits from here plus the Basic RP validation work. The fresh-context re-reviews of #321 and #322 found the same P1 issues (unconditional JWKS cache clearing in `conformance/app.py`, bare `assert`s), so splitting them into two review/merge cycles adds no value. Closing in favor of the combined PR.

PR #322, opened 2026-04-10
feat(conformance): OIDC RP harness + Basic RP certification (#242)

jamescrowley321:
## Summary

Combined PR covering:

1. **Conformance harness** (originally #321): FastAPI RP app in `conformance/`, Docker Compose with OIDF suite, automated test runner using the suite REST API
2. **Basic RP certification** (originally #322): library fixes driven by conformance test failures

Part of #242 (OIDC RP Certification). #321 closed as superseded — this branch chains from `feat/conformance-harness` and contains all of its commits plus the Basic RP validation work.

## Library Changes

- `parsers.py`: filter JWKS by `use`+`kty` for kid-absent JWTs (RFC 7517 §4.2)
- `token_validation.py`: always retry JWKS fetch on signature failure for key rotation

## Claimed Conformance Results

> ⚠️ **UNVERIFIED EVIDENCE.** The table below is hand-authored from source review. No conformance suite run artifacts (JSON export, log files, CI job) are committed to this PR. Fresh-context re-review flagged this as P1. Before treating Basic RP as certified, the suite must actually be run and its output committed as artifacts. See review comment for details.

| Test | Claimed | Notes |
|------|--------|-------|
| oidcc-client-test | PASSED | Basic authorization code flow end-to-end |
| oidcc-client-test-invalid-iss | PASSED | RP correctly rejects wrong issuer |
| oidcc-client-test-missing-sub | PASSED | RP correctly rejects missing sub claim |
| oidcc-client-test-invalid-aud | PASSED | RP correctly rejects wrong audience |
| oidcc-client-test-missing-iat | PASSED | RP correctly rejects missing iat claim |
| oidcc-client-test-kid-absent-single-jwks | PASSED | RP selects single signing key by use+kty filtering |
| oidcc-client-test-kid-absent-multiple-jwks | PASSED | RP correctly rejects (multiple signing keys, no kid) |
| oidcc-client-test-idtoken-sig-rs256 | PASSED | RP validates RS256 signature |
| oidcc-client-test-idtoken-sig-none | SKIPPED | Suite skips alg=none for client_secret_basic config |
| oidcc-client-test-invalid-sig-rs256 | PASSED | RP correctly rejects bad signature |
| oidcc-client-test-userinfo-invalid-sub | PASSED | RP detects UserInfo sub mismatch |
| oidcc-client-test-nonce-invalid | PASSED | RP correctly rejects wrong nonce |
| oidcc-client-test-scope-userinfo-claims | PASSED | RP requests standard scopes, gets UserInfo claims |
| oidcc-client-test-client-secret-basic | PASSED | RP uses client_secret_basic auth correctly |

**Claimed summary: 13 PASSED, 0 FAILED, 1 SKIPPED** — needs suite-run evidence to promote from "claimed" to "verified".

## Known Review Findings (P1 — should fix before merge)

1. `conformance/app.py` `clear_jwks_cache()` called unconditionally on every callback — defeats key-rotation test validation. The `rp-key-rotation-*` tests require the cached-key path to fail first, then retry. Pre-clearing skips the retry logic entirely.
2. Bare `assert` statements for None-checks in `conformance/app.py` (lines 251, 345, 370) — stripped under `python -O` / `-OO`.
3. No CI job runs the suite. Phase 2 of #242 explicitly lists CI integration as a deliverable.

## Unit Test Results
- Unit tests: 841 passed
- Coverage: 94.89%


conformance/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    networks:
+      default:
+        aliases:
+          - localhost.emobix.co.uk
+    depends_on:
+      server:
+        condition: service_healthy
+
+  server:
+    image: authelia/openid-connect-conformance-suite:latest
```
jamescrowley321, line 21: Explain your selection

jamescrowley321:
## Fresh-Context Adversarial Review (post-merge re-review)

### Blind Hunter

**MUST FIX**

- [`conformance/app.py:334`] `DiscoveryPolicy(require_https=False, validate_issuer=False)` is used **in the callback handler** for every token fetch, not just the initial discovery. Disabling issuer validation in the RP harness means the harness will never catch an issuer mismatch at the discovery layer. The library's own issuer validation (in `decode_with_config`) still fires, but the inconsistency between what the harness advertises and what the library enforces is a semantic gap. If the harness is what ran the conformance suite, it validated issuer in the ID token but not in the discovery document — so the `oidcc-client-test-invalid-iss` PASS may reflect library-level enforcement only, not full RP-level enforcement.

- [`conformance/app.py:384`] `clear_jwks_cache()` is called **unconditionally on every callback**, not just for key-rotation tests. This defeats the JWKS cache for every single test run, making the harness exercise only the uncached/first-fetch code path. Tests that claim to validate key rotation (`oidcc-client-test-kid-absent-single-jwks`, key rotation tests) are not actually testing the cache path because the cache is always cleared before use.

**SHOULD FIX**

- [`conformance/app.py:255`] `assert disco.authorization_endpoint is not None` — a bare `assert` in a FastAPI route will silently pass with `python -O` or produce an unhelpful 500 instead of a 502. Same pattern at lines 350–351. Use explicit `if` + error response instead.

- [`conformance/run_tests.py:898`] `module_id: str = module_info.get("id", module_info.get("name", "")) or ""` — triple-defaulting with an `or ""` suggests uncertainty about the API contract. If `id` is absent, the test module ID is silently empty and the subsequent poll will fail with a confusing 404. A missing `id` should be a hard error.

- [`conformance/app.py:182-195`] Global singleton HTTP client initialized via `global _http_client` — this pattern is flagged by ruff (PLW0603) and suppressed. The singleton is never closed: no lifespan handler, no `__del__`. In a long-running test session, this leaks connections.

**NITPICK**

- [`conformance/run_tests.py:710`] `SUITE_BASE_URL` is hardcoded, not read from environment. The `--suite-url` CLI arg allows override but the env var path used by the container (`CONFORMANCE_SUITE_BASE_URL`) is ignored by the runner script, only by the FastAPI app.

---

### Edge Case Hunter

| `conformance/app.py:409-420` | `id_token_jwt` is present but `claims` dict is empty (OP returns no `nonce` claim on purpose for some tests) | `if "nonce" in claims and claims["nonce"] != session.nonce:` | **[WRONG]** Nonce mismatch detection is correct for normal tokens but unclear for tests where OP intentionally omits nonce — current code treats absent nonce as `None != session.nonce`, which is a spurious failure |
| `conformance/app.py:308` | `sessions.pop(state_value)` — session popped before all validation succeeds | No change needed — pop is correct for CSRF, but if `_store_test_result` is never called on a code path, test result is unrecoverable | **[DEGRADED]** Test result not stored for some error paths if exception escapes the handler |
| `conformance/app.py:479-480` | `form_data` may contain multiple values per key (RFC 2388 multi-value fields) | `dict(form_data)` drops all but one value | **[WRONG]** Form POST callback with repeated parameters (unusual but conformance suite may send) silently drops values |
| `core/parsers.py:139` | `signing_keys` after `use` filter is non-empty but all have `use="enc"` and the fallback fires, then `kty_filtered` after alg filter is also non-empty — but the JWKS contains a mix of RSA enc + EC sig keys | Logic is correct — fallback only fires when signing_keys is empty; but the double-filter (`use` then `kty`) can yield a key with `use=None` and wrong `kty` being selected | **[WRONG]** If JWKS has one RSA-enc key and one EC-sig key, and alg is ES256: `signing_keys` = [EC-sig, RSA-enc... wait, use=enc is filtered], actually only EC-sig passes; then kty filter passes EC only — this is correct. No issue. |
| `aio/token_validation.py` (post-diff) | `_retry_with_refreshed_jwks` now always retries, even on non-cached paths. If the JWKS endpoint is down, the retry will raise `httpx.HTTPError` which is **not** caught here — the original `SignatureVerificationException` is swallowed and a network error propagates instead | `except (SignatureVerificationException, httpx.HTTPError):` in the caller | **[CRASH]** Upstream JWKS failure during retry surfaces as an undocumented network exception instead of a `TokenValidationException` |

**Summary**
- Unhandled paths found: 4

---

### Acceptance Auditor

The issue (#242) defines Phase 2 as the deliverable for this PR: Build the harness, run tests, report results. Extracting ACs from the phase structure:

**AC-1**: Build thin FastAPI harness using py-identity-model's public API
- **PASS** — `conformance/app.py` exists, uses the sync public API correctly. Tested implicitly by conformance runs.

**AC-2**: Docker Compose setup with OIDF conformance suite
- **PASS** — `conformance/docker-compose.yml` includes `authelia/openid-connect-conformance-suite:latest`, MongoDB, nginx TLS termination, RP harness. Correct architecture.

**AC-3**: Automated test runner using conformance suite REST API
- **PASS** — `conformance/run_tests.py` implements plan creation, module execution, polling, and result collection.

**AC-4**: CI integration running conformance tests
- **FAIL** — No CI workflow change is present in the diff. The conformance tests are **not** wired into CI. The PR description says "Unit tests: 841 passed, Coverage: 94.89%" but this is the standard test suite, not the OIDF suite. Issue #242 explicitly lists "CI integration — Run conformance tests in CI against the local suite" as a Phase 2 deliverable. This is unimplemented.

**AC-5**: All Basic RP tests pass (13 PASSED, 0 FAILED, 1 SKIPPED claimed)
- **PARTIAL** — The PR body contains a results table but **no artifact proves it**. No log file, no CI run output, no exported JSON from the suite. The test runner has the machinery to collect results, but no output is committed. This is a hand-authored table. Per issue #242's own warning: "a merged PR is necessary but not sufficient evidence that a conformance test passes." See Evidence Check below.

**AC-6**: Fix kid-absent key selection (use + kty filtering per RFC 7517 §4.2)
- **PASS** — `find_key_by_kid` and `get_public_key_from_jwk` both updated with the two-stage filter. Unit tests added and correct.

**AC-7**: Always retry JWKS on signature failure (key rotation)
- **PASS** — `_retry_with_refreshed_jwks` now fires regardless of whether the path was cached. However, see Edge Case Hunter: unhandled HTTP errors during retry.

**SCOPE CREEP**

- `sonar-project.properties` adds `s5659_parsers` suppression for `get_unverified_header`. This is a SonarCloud quality gate workaround. It's reasonable (the behavior is intentional), but it wasn't in any AC.
- `conformance/` is excluded from SonarCloud analysis entirely. Appropriate given it's test infra, but this was a judgment call not in the issue.

**Summary**: Total ACs: 7 | Pass: 5 | Fail: 1 (CI integration) | Partial: 1 (conformance run evidence)

---

### Sentinel (Security)

**BLOCK (must fix before merge)**

- No findings that apply to the library itself. The library changes (parsers.py, token_validation.py) increase strictness: use+kty filtering for kid-absent selection is correct per RFC 7517 §4.2, and always retrying on signature failure is correct per OIDC Core §10.1.5. The `validate_issuer=False` and `require_https=False` in `conformance/app.py` are scoped to the test harness only, which is excluded from SonarCloud and the library distribution.

**WARN**

- [LIKELY] `conformance/app.py:194` — `HTTPClient(verify=False)` is a process-global singleton. If someone imports `conformance/app.py` outside the Docker container context (e.g., in a test script running against a real OP), TLS verification is silently disabled. This is test infrastructure so the risk is contained, but the global nature means there's no per-request opt-out.

- [UNLIKELY] `conformance/app.py:182-185` — `sessions` and `test_results` are plain Python dicts with no size bound. A conformance suite that creates thousands of test modules without completing them would grow these dicts without bound. In practice the suite is controlled, so this is not exploitable in normal use.

**INFO**

- The library change removing `if not is_cached: raise` in the signature failure handler (both sync and async) means a genuinely bad token (wrong issuer's key, not a rotation event) will now trigger a second JWKS fetch before raising. This adds one extra HTTP call on every permanently-invalid token. Not a security issue, but a mild denial-of-service amplification if callers pass attacker-controlled JWTs in a tight loop.

**Summary**: BLOCK: 0 | WARN: 2 | INFO: 1 | Overall: PASS

---

### Viper (Red Team)

**Attack Surface**

- `find_key_by_kid` key selection logic changed — new code path for kid-absent multi-key JWKS
- JWKS retry logic changed — retry now unconditional (was: only on cached path)
- `conformance/app.py` — new unauthenticated RP endpoints (intended for test harness)

**Findings**

#### LOW — Infinite retry amplification via attacker-controlled JWT

- **Location**: `sync/token_validation.py:204`, `aio/token_validation.py:207`
- **Attack scenario**:
  1. Attacker submits a JWT signed by a key not in the OP's JWKS (e.g., self-signed)
  2. Library fetches JWKS, fails signature check
  3. Library re-fetches JWKS unconditionally (new behavior), fails again
  4. JWKS endpoint receives two requests per bad token submission
  5. At scale: N bad tokens = 2N JWKS fetches against the OP
- **Prerequisites**: Ability to submit JWTs to a py-identity-model-protected endpoint
- **Remediation**: Add a jitter/backoff between retries; or limit retry to cases where the kid changed between first and second JWKS fetch

#### LOW — kid-absent fallback is attacker-influenceable via JWT `alg` header

- **Location**: `core/parsers.py:141-147`
- **Attack scenario**:
  1. JWKS contains one RSA signing key and one EC signing key
  2. Attacker crafts a JWT without `kid` and sets `alg=ES256`
  3. Library filters to EC key via kty match and uses it for verification
  4. If attacker also controls the EC key material (in a confused deputy scenario), the wrong key is selected
- **Prerequisites**: Attacker must control JWT header AND control the EC key registered in the JWKS — this requires a compromised OP or a key confusion scenario, making it very low likelihood
- **CVSS v3.1**: 2.6 (AV:N/AC:H/PR:H/UI:N/S:U/C:L/I:N/A:N)
- **Remediation**: This is the spec-correct behavior (RFC 7517 §4.2 + alg narrowing). Document that callers should ensure their JWKS does not have multiple signing keys with distinct kty unless the OP is trusted to set `kid`. No code change needed.

**Summary**: Attack surface elements: 3 | Findings: 0 critical, 0 high, 0 medium, 2 low | Overall: PASS

---

### Triage summary

- **P0**: 0
- **P1**: 2
  - Unconditional `clear_jwks_cache()` on every callback means the harness never exercised the JWKS cache code path during conformance runs. Tests that claim to validate key rotation behavior may have passed for the wrong reason.
  - No actual conformance run output committed. The 13 PASSED table in the PR body is asserted, not evidenced. If this PR is the basis for submitting to `certification.openid.net`, that submission would be fraudulent without real run artifacts.
- **P2**: 3
  - CI integration not wired (Phase 2 AC explicitly listed in issue, not implemented)
  - Unconditional JWKS retry on non-cached path — HTTP error during retry swallows original `SignatureVerificationException` and surfaces as undocumented network exception
  - `validate_issuer=False` in harness callback means discovery-layer issuer validation is disabled in the RP harness (library-level token issuer check still applies)

**Overall gate**: ⚠️ MERGE WITH FOLLOWUP

The library changes (parsers.py, token_validation.py) are correct, spec-compliant, and well-tested. They make the library strictly more correct. The conformance harness infrastructure is sound in structure. The issues above are in the harness and the evidence layer, not the library itself.

---

### Evidence check

- **Does this PR include actual conformance test run output?** NO
- The PR body contains a hand-authored results table with "13 PASSED, 0 FAILED, 1 SKIPPED" but zero artifacts: no exported JSON, no log file, no screenshot of the conformance suite UI, no CI workflow that runs the suite.
- The CI checks all pass but they run the **standard test suite** (unit, integration, examples) — not the OIDF conformance suite. There is no `conformance` CI job.
- The issue (#242) explicitly warned: "a merged PR is necessary but not sufficient evidence that a conformance test passes." This PR has not met the sufficient-evidence bar.
- The run_tests.py script has the machinery to output results, but no captured output was committed alongside the harness.
- **Verdict**: The conformance results table must be treated as unverified. A follow-up should commit actual run output (the runner's stdout or an exported suite plan JSON) before the library is submitted to certification.openid.net.

jamescrowley321:
Superseded by #325, which inherits all commits from this PR's branch via the `feat/conformance-harness` → `feat/basic-rp-conformance` → `ci/conformance-suite-in-ci` → `feat/config-rp-conformance` chain. Merging #325 lands this PR's commits on main. Closing to consolidate the merge into a single operation per owner directive.

PR #340, opened 2026-04-11
feat(conformance): add Playwright-driven token rotation script for HCP Vault

jamescrowley321:
## Summary

Adds \`conformance/scripts/rotate_conformance_token.py\` — a standalone script that automates the OIDF certification suite API token lifecycle:

1. Launches Chromium with a persistent user-data profile
2. Navigates to https://www.certification.openid.net/
3. Waits for the operator to complete the interactive OIDC login (Google or GitLab) on first run; reuses the persistent session on subsequent runs
4. Navigates to the token management page and creates a new API token via the UI
5. Captures the token value and pushes it to HCP Vault Secrets via the \`hcp\` CLI

## Why

The hosted runner work in PR #338 needs a Bearer token to call \`certification.openid.net\`'s REST API. The token creation endpoint requires an OIDC-authenticated browser session, which cannot be fully automated headlessly (Google actively blocks headless Chromium on login). This script gets us a reproducible way to rotate that secret without hand-editing \`.env\` files or pasting into CI secrets manually.

Related: #242, unblocks #338.

## Design highlights

- **PEP 723 inline dependencies** — runnable as \`uv run conformance/scripts/rotate_conformance_token.py\` with no project-level dep changes. Playwright is pulled into an ephemeral venv on demand.
- **Persistent browser profile** over fighting Google's bot detection. First run is headful + manual sign-in; cookies stick in \`~/.cache/py-identity-model/playwright-profile/\`; later runs can be headless.
- **UI-driven token creation** rather than hitting \`POST /api/token\` directly. The suite's REST endpoint exists (\`TokenApi.java\`) but its request/response format isn't publicly documented; the UI is the stable, documented interface. The module docstring points future maintainers at the REST fallback path.
- **hcp CLI for the secrets push.** Token passed via \`--data-file=-\` (stdin) so it never appears in argv visible to \`ps\`. CLI path is resolved via \`shutil.which\` so we pass an absolute argv[0].
- **No passwords in the script.** Operator credentials never touch environment variables — all auth happens in the browser window.
- **Safe defaults.** Masked output, \`--dry-run\` flag to verify the browser flow without touching Vault, \`--show-token\` opt-in for the full value.

## Flags

| Flag | Purpose |
|---|---|
| \`--headless\` | Run Chromium headless (only after first interactive run has cached session cookies) |
| \`--dry-run\` | Create the token but don't push to Vault |
| \`--show-token\` | Print the full token value (default is masked) |
| \`--profile-dir\` | Override the persistent profile location |
| \`--app-name\` | HCP Vault Secrets app name (default: \`py-identity-model\`) |
| \`--secret-name\` | Secret name inside the app (default: \`CONFORMANCE_TOKEN\`) |
| \`--description\` | Description attached to the new token in the suite UI |

## Test plan

- [x] \`ruff check\` + \`ruff format\` pass
- [x] \`pyrefly check\` passes
- [x] \`py_compile\` passes
- [ ] Manual run against certification.openid.net with a real Google account (requires owner)
- [ ] Manual verification that the resulting secret appears in HCP Vault Secrets under \`py-identity-model/CONFORMANCE_TOKEN\` (requires owner)

The Playwright selectors for the token management page are a first-draft best guess based on the suite's upstream template. The module docstring explicitly notes that if the UI layout changes the selectors need to be updated — and documents the REST fallback for that case.

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -42,3 +42,34 @@ The conformance suite uses `localhost.emobix.co.uk` which resolves to `127.0.0.1
 | `GET /callback` | GET | Handle authorization callback |
 | `POST /callback` | POST | Handle form_post callback |
 | `GET /results/{test_id}` | GET | Retrieve test flow results |
+
+## Hosted certification token rotation
+
+For cert-grade runs against `https://www.certification.openid.net/` (the hosted
+OIDF suite) the runner needs a Bearer token. The token is created by an
+interactive browser session against the hosted suite — there is no headless
+API to create one because the creation endpoint itself requires an
+OIDC-authenticated browser session.
+
+`scripts/rotate_conformance_token.py` automates the non-interactive parts of
+that flow: persistent browser profile, token creation via the suite's UI,
+and pushing the resulting secret to HCP Vault Secrets.
+
+```bash
+# First run — interactive Google/GitLab sign-in in the browser window
+uv run conformance/scripts/rotate_conformance_token.py
+
+# Subsequent runs — persistent profile keeps you signed in
+uv run conformance/scripts/rotate_conformance_token.py --headless
+
+# Dry run — create the token but print (masked) instead of pushing to Vault
+uv run conformance/scripts/rotate_conformance_token.py --dry-run
+```
+
+**Prerequisites:**
+- `uv` (PEP 723 inline dependency support)
+- Playwright Chromium binary: `uv run --with playwright playwright install chromium`
+- HCP CLI installed and authenticated: `hcp auth login` + `hcp profile init`
+- HCP Vault Secrets app configured (default name: `py-identity-model`)
+
+See the script's module docstring for full design notes and flag reference.
--- a/conformance/scripts/rotate_conformance_token.py
+++ b/conformance/scripts/rotate_conformance_token.py
@@ -1,0 +1,466 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.10"
+# dependencies = ["playwright>=1.40"]
+# ///
+"""Rotate the OIDC conformance-suite API token and push it to HCP Vault Secrets.
+
+Target service: https://www.certification.openid.net/
+
+Why this exists
+---------------
+The OIDF certification server exposes a REST API (`/api/plan`, `/api/runner`,
+`/api/info`, `/api/plan/{id}/certificationpackage`) that is authenticated via a
+Bearer token. Tokens are created via a single interactive step: log in to the
+suite via Google or GitLab OIDC in a browser, navigate to the API token
+management page, and click "Create". Everything after that (running the test
+plans, publishing the certification package) is plain HTTP against the REST
+API using the Bearer token.
+
+This script automates the token creation + storage half of that workflow so
+token rotation is reproducible and the resulting secret lives in HCP Vault
+Secrets instead of a `.env` file on someone's laptop.
+
+Flow
+----
+1. Launch Chromium with a persistent user-data dir (so OIDC sessions stick).
+2. Navigate to https://www.certification.openid.net/.
+3. If not logged in, wait for the operator to complete the OIDC flow
+   interactively (Google/GitLab). First run is always headful.
+4. Navigate to the token management page and create a new API token.
+5. Capture the token value from the UI.
+6. Push the value to HCP Vault Secrets via `hcp vault-secrets secrets create`.
+
+Prerequisites
+-------------
+- Python 3.10+
+- uv (https://docs.astral.sh/uv/) for the PEP 723 inline script runner
+- Playwright Chromium browser binary:
+      uv run --with playwright playwright install chromium
+- HCP CLI (https://developer.hashicorp.com/hcp/docs/cli/install) already
+  authenticated via `hcp auth login` and scoped to the target org/project
+  via `hcp profile init`.
+
+Usage
+-----
+First run (interactive OIDC login — browser opens, you sign in manually):
+
+      uv run conformance/scripts/rotate_conformance_token.py
+
+Subsequent runs (persistent profile keeps you logged in, can be headless):
+
+      uv run conformance/scripts/rotate_conformance_token.py --headless
+
+Preview what the script will do without pushing to Vault:
+
+      uv run conformance/scripts/rotate_conformance_token.py --dry-run
+
+Show the full token value on stdout instead of masking it (use with care;
+anything that stores stdout persistently — shell history, CI logs — becomes
+a secret spill surface):
+
+      uv run conformance/scripts/rotate_conformance_token.py --show-token
+
+Environment variables
+---------------------
+HCP_VAULT_APP_NAME     HCP Vault Secrets app to push to
+                       (default: py-identity-model)
+HCP_VAULT_SECRET_NAME  Secret name inside the app
+                       (default: CONFORMANCE_TOKEN)
+PLAYWRIGHT_PROFILE_DIR Override the persistent browser profile directory
+                       (default: ~/.cache/py-identity-model/playwright-profile)
+
+Design notes
+------------
+- **Persistent profile over headless auth**: Google and GitLab both actively
+  block headless Chromium on login pages (UA detection + "this browser or app
+  may not be secure" blocks). Letting the operator sign in once in a headful
+  session and reusing the cookie jar is far more reliable than trying to
+  defeat the bot detection. First run is interactive; subsequent runs can
+  be headless as long as the session cookie is still valid.
+
+- **UI-driven token creation**: The suite has a REST endpoint at POST
+  /api/token but its exact request/response format is not publicly
+  documented outside the `TokenApi.java` source. Rather than guess, this
+  script drives the suite's UI (which is a stable, documented interface).
+  If the UI layout changes, fix the selectors in `_create_api_token_via_ui`
+  and rerun.
+
+- **Push via the hcp CLI, not the REST API**: HCP Vault Secrets has an HTTP
+  API, but the `hcp` CLI is the easier surface — it handles auth token
+  refresh, org/project scoping, and error reporting. Shelling out is fine
+  for a script that runs occasionally under human supervision.
+
+- **No credentials in the script**: The script never reads Google/GitLab
+  passwords from environment variables. All user auth happens in the
+  browser the operator controls.
+"""
+
+from __future__ import annotations
+
+import argparse
+from dataclasses import dataclass
+import os
+from pathlib import Path
+import shutil
+import subprocess
+import sys
+
+from playwright.sync_api import Page, sync_playwright
+from playwright.sync_api import TimeoutError as PlaywrightTimeoutError
+
+
+SUITE_URL = "https://www.certification.openid.net/"
+TOKEN_MANAGEMENT_PATH = "token-management.html"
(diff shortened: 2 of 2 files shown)


jamescrowley321:
## Fresh-Context Adversarial Re-Review — PR #340

This is an AI-authored PR for a security-sensitive script (handles a Bearer token to OIDF's cert suite and pushes to HashiCorp Vault). Owner requested fresh-context adversarial review.

---

### Blind Hunter

**MUST FIX**

No findings.

**SHOULD FIX**

- [`rotate_conformance_token.py:305`] `submit.first.click()` has no guard — if the "Create" button (distinct from "Create API Token"/"New Token") is missing from the dialog, Playwright will raise `TimeoutError` after 30 s with a message like "Timeout 30000ms exceeded waiting for locator...". That IS caught in `main()`, but the error message gives no hint that the Create _submission_ button is the problem (unlike the earlier guard at line 333 which raises a helpful `RuntimeError`). A user seeing this failure won't know which selector to fix.

- [`rotate_conformance_token.py:407`] `_ensure_hcp_cli_available()` runs `hcp --version` with `check=True`, but `capture_output=True` means stderr is captured. The `except SubprocessCalledProcessError` re-raises as `RuntimeError` using `exc.stderr.decode()`. If `hcp` is installed but crashes during `--version` (e.g., corrupt binary or missing shared library), the decode may fail if the process wrote non-UTF-8 bytes to stderr. Should use `exc.stderr.decode("utf-8", errors="replace")`.

**NITPICK**

- [`rotate_conformance_token.py:219`] The argparse `--profile-dir` argument has `default=None` and the actual resolution happens in `build_config()`. The `help` text manually embeds the default path string (`{DEFAULT_PROFILE_DIR}`). If `DEFAULT_PROFILE_DIR` ever changes, the help text silently diverges. Prefer passing the resolved path as `argparse` default directly, or at minimum add a comment noting the intentional split.

---

### Edge Case Hunter

| `rotate_conformance_token.py:305` | "Create" dialog button absent after UI interaction (selector changed) | `if submit.count() == 0: raise RuntimeError("Could not find 'Create' submit button...")` | **[DEGRADED]** 30-second hang then opaque `TimeoutError` with no actionable hint |
| `rotate_conformance_token.py:273` | `page.goto(SUITE_URL)` with `wait_until="networkidle"` — page default timeout is 30 s; certification.openid.net is sometimes slow | already uses `set_default_timeout` but `goto` has its own 30 s default that overrides it | **[DEGRADED]** Silent timeout with no retry; user must re-run manually |
| `rotate_conformance_token.py:253-255` | `_needs_login` returns `False` when the session cookie is stale (server-side expired but cookie still present) — the Login link may be absent even when the session is dead | No guard; detected only downstream when token management page redirects or returns 401 | **[DEGRADED]** Confusing failure: script thinks it's logged in, navigates to token-management.html, then fails with a selector-not-found Playwright error |
| `rotate_conformance_token.py:407`| `hcp --version` stderr contains non-UTF-8 bytes (corrupt binary or locale issue) | `exc.stderr.decode("utf-8", errors="replace")` | **[CRASH]** `UnicodeDecodeError` escapes `except CalledProcessError`, propagates to `main()` as unhandled, crashes with traceback |

**Summary**
- Unhandled paths found: 4
- Critical (crash): 1 (`UnicodeDecodeError` in `_ensure_hcp_cli_available`)
- Non-critical (degraded/wrong): 3

---

### Acceptance Auditor

There is no linked GitHub issue. Per the review instructions, I audit against the PR description's own claimed scope.

**Extracting ACs from PR description:**

- AC-1: Launch Chromium with a persistent user-data profile
- AC-2: Navigate to https://www.certification.openid.net/
- AC-3: Wait for interactive OIDC login on first run; reuse session on subsequent runs
- AC-4: Navigate to token management page and create a new API token via UI
- AC-5: Capture token and push to HCP Vault Secrets via `hcp` CLI
- AC-6 (Design): PEP 723 inline dependencies — runnable via `uv run` with no project-level dep changes
- AC-7 (Design): Persistent browser profile under `~/.cache/py-identity-model/playwright-profile/`; subsequent runs can be headless
- AC-8 (Design): Token passed via `--data-file=-` (stdin), never in argv
- AC-9 (Design): Masked output by default; `--show-token` opt-in; `--dry-run` flag
- AC-10 (Design): No credentials in the script
- AC-11: README section documenting workflow and prerequisites

**PASS**

- [AC-1] Implemented at `rotate_conformance_token.py:264-269` (`launch_persistent_context`)
- [AC-2] Implemented at `rotate_conformance_token.py:273` (`page.goto(SUITE_URL)`)
- [AC-3] Implemented at `rotate_conformance_token.py:275-289` (`_needs_login` + `_wait_until_logged_in`)
- [AC-4] Implemented at `rotate_conformance_token.py:290`, `_create_api_token_via_ui`
- [AC-5] Implemented at `rotate_conformance_token.py:437`, `push_to_hcp_vault_secrets`
- [AC-6] PEP 723 block at lines 1-4 is syntactically correct; no project-level changes in diff
- [AC-7] Profile dir at `DEFAULT_PROFILE_DIR`; `--headless` guard for unauthenticated state
- [AC-8] `--data-file=-` with `input=token` via `subprocess.run`; confirmed token never in `cmd` list
- [AC-10] No credential env vars read anywhere in the script
- [AC-11] README section added at `conformance/README.md` lines 10-39

**PARTIAL**

- [AC-9] Masked output and `--dry-run` are correctly implemented. However, the module docstring (line 58) states `--show-token` prints "on **stdout**", while the code at line 425 always writes to **stderr** (`file=sys.stderr`). The code is safer, but the documented behavior is wrong. This also means that `--show-token` cannot be used to pipe the token value to another command (`... | xargs`) — the docstring implies it can.

**FAIL**

No implemented ACs are missing; no scope creep.

**Tests:** The PR test plan explicitly marks the only automated checks (ruff, pyrefly, py_compile) as passing and marks the two functional verifications as requiring manual owner action. This is the right call for a Playwright script — no automated tests exist and none are claimed.

**Summary**
- Total ACs: 11
- Pass: 10 | Fail: 0 | Partial: 1

---

### Sentinel (Security)

#### BLOCK (must fix before merge)

No BLOCKs.

#### WARN (should fix)

- [LIKELY] `rotate_conformance_token.py:370-373` — **HCP CLI stdout/stderr leaked in error messages without sanitization.** When `hcp vault-secrets secrets create` fails, the `RuntimeError` includes `result.stderr` and `result.stdout` verbatim. Some HCP CLI error responses echo back request context. More critically, if a previous secret version with the same name is being overwritten and the CLI emits it in the error body (unlikely but not impossible with some HCP versions), the old token value would appear in the error string, which is then printed to stderr via `main()`. This is low-probability but the blast radius is a Bearer token being logged to a terminal session's scrollback buffer.
  - Mitigation: Sanitize the error output — at minimum, note in a comment that `result.stdout` is included and that the token value should not appear there. Stronger fix: cap the length of stderr/stdout included in the error message.

- [UNLIKELY] `rotate_conformance_token.py:295-303` — **Stale-session false-negative in `_needs_login`.** If a session cookie is present but server-side expired, the Login link is absent and `_needs_login` returns `False`. The script proceeds to `token-management.html`, which will redirect to login. This redirect may (depending on the conformance server's behavior) result in a Playwright selector failure rather than a clear "session expired" message. An attacker cannot exploit this directly, but an operator in a shared environment might share a profile dir and unknowingly operate on someone else's session cookies.
  - Mitigation: After navigating to the main page, confirm presence of a post-login element (e.g., user display name or logout link) rather than only checking absence of a Login link.

#### INFO (acceptable risk)

- `rotate_conformance_token.py:264-269` — **Browser profile directory created with default umask permissions.** `cfg.profile_dir.mkdir(parents=True, exist_ok=True)` inherits the process umask; on most systems this is `0o755` (world-readable directory). The profile dir stores Google/GitLab OIDC session cookies. On a single-user development machine this is acceptable. On a shared system (e.g., a CI runner), other users could potentially read the session cookies. The PR description does not claim to set restrictive permissions. Acceptable for a dev-only tool, but worth documenting.

- `rotate_conformance_token.py:397-411` — **`_ensure_hcp_cli_available` does NOT verify HCP authentication.** `hcp --version` succeeds even when `hcp auth login` has not been run. Authentication failure surfaces only at the actual `secrets create` call. This matches the docstring's stated design intent. The error from `hcp` at that point will be actionable. Acceptable.

- `rotate_conformance_token.py:413-415` — **Token value persists in Python heap as an immutable string** until GC. No zeroing attempt. For a developer laptop tool this is acceptable — the threat model doesn't include memory forensics. Noted as defense-in-depth observation only.

- `rotate_conformance_token.py:397-411` — **`hcp` binary resolved via `shutil.which` at startup and the resolved path is passed to all subsequent calls.** This prevents PATH-based binary substitution attacks between the validation call and the secrets call. Good practice, correctly implemented.

- `rotate_conformance_token.py:351-363` — **TLS verification:** Playwright uses Chromium's default TLS stack. No `ignore_https_errors=True` is passed anywhere in the `launch_persistent_context` call. TLS verification for `certification.openid.net` is enforced.

- `rotate_conformance_token.py:352-365` — **No argv injection from env vars.** `app_name` and `secret_name` come from env vars or CLI args but are passed as discrete list elements in `cmd` (not interpolated into a shell string, `shell=False` by default). A value like `"foo --flag=evil"` in `HCP_VAULT_APP_NAME` would be passed as a single argument to the HCP CLI — the CLI would reject it, not interpret the embedded flag. No injection possible.

**Summary**
- BLOCK: 0 | WARN: 2 | INFO: 6
- Overall: PASS (no blocking security issues)

---

### Security checklist

- [x] Token leakage vectors audited — token goes to stderr (masked by default); `--show-token` sends to stderr not stdout (docstring is wrong but code is safer); token passed via stdin to HCP; HCP error path includes raw stdout/stderr (WARN above); no argv exposure
- [ ] Browser profile dir permissions — `mkdir` uses process umask; not explicitly hardened to `0o700`; acceptable for dev tool, not documented
- [x] Subprocess argv injection prevented — list-form subprocess, no `shell=True`, resolved absolute binary path
- [x] TLS verification enforced — no `ignore_https_errors` in Playwright launch args
- [x] HCP CLI error paths give actionable messages — yes for missing CLI, unauthenticated state surfaces at secrets-create call with HCP's own error

---

### Triage summary

- **P0 (blocks merge)**: 0
- **P1 (should fix)**: 3 — submit button missing guard (Blind Hunter SHOULD FIX), `UnicodeDecodeError` crash in `_ensure_hcp_cli_available` (Edge Case Hunter), `--show-token` stdout/stderr doc/code mismatch (Acceptance Auditor PARTIAL)
- **P2 (nitpick)**: 1 — argparse default display inconsistency

**Does the PR description match what's in the diff?** PARTIAL — 10/11 ACs pass. One discrepancy: docstring says `--show-token` prints to stdout; code sends it to stderr. The code behavior is correct and safer; the documentation is wrong.

**Overall gate: ⚠️ MERGE WITH FOLLOWUP**

The script is well-structured and the security-sensitive path (stdin handoff, masked defaults, argv list form, TLS verification, absolute binary path) is handled correctly. No blocking security issues. The three P1 items are all fixable in a quick follow-up: the submit-button missing guard is a UX regression risk if the OIDF UI changes; the `UnicodeDecodeError` is a real crash path; and the stdout/stderr docstring mismatch could mislead future callers who try to pipe `--show-token` output. Recommend tagging a follow-up issue and merging.

PR #363, opened 2026-04-12
fix(jwt): remove lru_cache from JWT decode to prevent stale expired tokens

### LINKED ISSUE #352: security: high — JWT decode cache returns stale results for expired tokens
## Finding H-4: JWT decode cache returns stale results for expired tokens

**Severity:** High
**Category:** Stale token acceptance
**Location:** `core/jwt_helpers.py:49-93`

### Description

`_decode_jwt_cached` uses `@lru_cache(maxsize=256)`. A token valid at cache time continues to be returned as valid after expiration until evicted. With 256 entries, a token could remain cached for minutes to hours.

### Analysis

LRU eviction is based on access frequency, not time. A frequently-used token that was valid when first decoded will continue to be accepted by the cache even after its `exp` claim has passed. This effectively extends the validity window of tokens beyond their intended lifetime, which is particularly dangerous for short-lived access tokens intended to limit the blast radius of token compromise.

### Suggested Fix

Remove JWT decode caching, or validate `exp` after cache retrieval. Note that `_get_pyjwk` already caches the expensive key construction operation, so the cache on JWT decode may provide limited benefit anyway.

---

Decomposed from #300

jamescrowley321:
## Summary

- **Remove `@lru_cache(maxsize=256)` from `_decode_jwt_cached`** to fix a security issue where expired JWT tokens could be returned from cache instead of being re-validated on each call
- **Rename `_decode_jwt_cached` to `_decode_jwt`** and simplify its signature to accept native types (list, dict) directly instead of hashable conversions (tuple, JSON string)
- **Remove the `.copy()` call** on the decoded dict return value — the cache aliasing concern only existed because `lru_cache` returns the same dict reference on cache hits
- **Keep `_get_pyjwk` lru_cache (maxsize=128)** — PyJWK construction is genuinely expensive (cryptographic key parsing) and keys don't expire

## Security Impact

The `lru_cache` on JWT decode meant that once a token was decoded and cached, subsequent validations of the same token would return the cached result without re-checking expiration. This allowed expired tokens to pass validation if they had been successfully validated before expiring. Removing the cache ensures every token validation performs a fresh decode with full expiration checking.

## Test plan

- [x] Updated mock targets in `test_jwt_helpers.py` from `_decode_jwt_cached` to `_decode_jwt`
- [x] Removed `_decode_jwt_cached.cache_clear()` calls from fixtures in `test_aio_token_validation.py`, `test_sync_token_validation.py`, and `test_enhanced_token_validation.py`
- [x] Removed `_decode_jwt_cached` import from `test_enhanced_token_validation.py`
- [x] All 848 unit tests pass with 95.21% coverage
- [x] `make lint` passes (ruff, ruff format, pyrefly, coverage)

Closes #352

### DIFF
--- a/src/py_identity_model/core/jwt_helpers.py
+++ b/src/py_identity_model/core/jwt_helpers.py
@@ -46,51 +46,45 @@ def _get_pyjwk(key_json: str, algorithm: str | None) -> PyJWK:
     return PyJWK(json.loads(key_json), algorithm)
 
 
-@lru_cache(maxsize=256)
-def _decode_jwt_cached(  # noqa: PLR0913  # @lru_cache requires individual hashable params
-    jwt: str,
-    key_json: str,
-    algorithms_tuple: tuple[str, ...],
+def _decode_jwt(  # noqa: PLR0913  # JWT validation requires these params per RFC 7519
+    jwt_token: str,
+    key: dict,
+    algorithms: list[str],
     audience: str | None,
-    issuer: str | tuple[str, ...] | None,
-    options_json: str | None,
+    issuer: str | list[str] | None,
+    options: dict | None,
     leeway: float | None = None,
 ) -> dict:
     """
-    Internal cached JWT decoding.
+    Internal JWT decoding.
 
-    Caches decoded tokens to avoid redundant signature verification when
-    the same JWT is validated multiple times.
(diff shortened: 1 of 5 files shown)


jamescrowley321:
Superseded by #364 which includes this change plus the full caching overhaul (discovery TTL cache, PyJWK cache removal, cross-user isolation tests).

Issue #353, opened 2026-04-12
security: medium — no JWKS response size limit enables DoS via oversized responses


jamescrowley321:
## Finding M-1: No JWKS response size limit → DoS via oversized responses

**Severity:** Medium
**Category:** Denial of Service
**Location:** `core/response_processors.py:278-280`

### Description

Processes unlimited keys from JWKS responses. An attacker-controlled JWKS endpoint could return thousands of keys causing memory/CPU exhaustion.

### Analysis

Without a bound on the number of keys processed from a JWKS response, a malicious or compromised JWKS endpoint can cause the service to allocate unbounded memory and spend unbounded CPU time constructing key objects. This is a defense-in-depth concern particularly relevant in environments where the JWKS URI may not be fully trusted.

### Suggested Fix

Limit processed keys (e.g., max 100) and validate response size before parsing.

---

Decomposed from #300

jamescrowley321:
## Implementation Plan

**Batch:** 5 — Resource limits + concurrency (`fix/resource-limits`)

### Fix
- Add a configurable max response size for JWKS fetches (default: 512KB)
- Check `Content-Length` header before reading body; if absent, use streaming read with size tracking
- Reject responses exceeding the limit with a clear error
- Configurable via `JWKS_MAX_RESPONSE_SIZE` env var or `SecurityPolicy` field
- Apply to both sync and async HTTP clients

### Testing
- Test that a JWKS response under the limit is accepted
- Test that a response over the limit is rejected
- Test that the limit is configurable

### Risk
- 512KB default is generous — typical JWKS responses are <10KB. Unlikely to break real providers.

PR #384, opened 2026-04-15
fix(security): add JWKS max key count limit and KeyError guard

### LINKED ISSUE #376: security: high — no JWKS response size limit enables memory/CPU exhaustion
## Summary

`parse_jwks_response()` in `core/response_processors.py:311-312` processes the entire JWKS response without any size limit. No Content-Length check, no max key count, no response body size cap.

**Status:** Original finding M-1 from #300. #353 was closed but no size limit was implemented — grep for `Content-Length`, `max_jwks`, `size_limit`, or `max_response` returns no matches in the codebase.

## Affected Code

```python
# response_processors.py:311-312
response_json = response.json()   # Parses unlimited response body into memory
keys = [jwks_from_dict(key) for key in response_json["keys"]]  # Processes all keys
```

## Exploit Scenario

An attacker controlling a JWKS endpoint (directly or via a compromised discovery document pointing `jwks_uri` to a malicious URL) returns:
- A multi-megabyte response with thousands of keys → memory exhaustion
- Hundreds of complex RSA keys → CPU exhaustion during `jwks_from_dict()` → `_validate_key_parameters()` including modulus size calculation via `_decode_base64url`

## Recommended Fix

1. Check `Content-Length` header and reject responses > 512KB before calling `response.json()`
2. For chunked/streamed responses without `Content-Length`, use `response.read()` with a max size parameter
3. Limit processed keys to a configurable maximum (default 100)
4. Add test with 1000+ key JWKS response

## Related

- Original finding: #300 (M-1)
- Prior issue: #353 (closed without implementation)

jamescrowley321:
## Summary

- Add max key count limit (default 100, configurable via `MAX_JWKS_KEYS` env var) to `parse_jwks_response()` to prevent CPU/memory exhaustion from oversized JWKS responses
- Guard `response_json["keys"]` access with proper KeyError handling instead of letting raw KeyError propagate
- Add TypeError guard for non-dict JWKS responses (e.g., JSON arrays)
- Validate individual key entries are dicts before processing
- Handle malformed Content-Length headers gracefully with warning log

Closes #376

## Exploit Scenario Blocked

An attacker controlling a JWKS endpoint returns a large JWKS JSON with 1000+ keys that stays under the 512KB body size limit. Each key triggers `jwks_from_dict()` → `_validate_key_parameters()` → CPU exhaustion on modulus calculation. The new key count limit (default 100) rejects such responses before processing any keys.

Additionally, a response missing the `"keys"` field entirely no longer raises a raw `KeyError` — it returns a descriptive error response.

## Review Findings Addressed

5-reviewer security review completed (Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, Viper):
- **3 P0 fixes**: Content-Length pre-check ordering, TypeError guard for non-dict responses, MAX_JWKS_KEYS clamped to min 1
- **3 P1 fixes**: Malformed Content-Length handling, per-element dict validation, warning logging
- **P2 deferred**: env var ValueError (consistent with existing codebase patterns)
- All 5 acceptance criteria passed
- Delta re-review confirmed all fixes correct, no new issues

## Test plan
- [x] Unit tests pass (`make test-unit`)
- [x] Lint passes (`make lint`)
- [x] Independent review agents passed (5/5)
- [ ] CI passes

### DIFF
--- a/src/py_identity_model/core/http_utils.py
+++ b/src/py_identity_model/core/http_utils.py
@@ -21,6 +21,7 @@ DEFAULT_HTTP_TIMEOUT = 30.0
 DEFAULT_RETRY_MAX_ATTEMPTS = 3
 DEFAULT_RETRY_BASE_DELAY = 1.0
 DEFAULT_MAX_JWKS_SIZE = 512 * 1024  # 512 KB
+DEFAULT_MAX_JWKS_KEYS = 100
 
 # HTTP status codes for retry logic
 HTTP_TOO_MANY_REQUESTS = 429
@@ -113,8 +114,18 @@ def get_max_jwks_size() -> int:
     return int(os.getenv("MAX_JWKS_SIZE", str(DEFAULT_MAX_JWKS_SIZE)))
 
 
+def get_max_jwks_keys() -> int:
+    """Get maximum number of keys allowed in a JWKS response.
+
+    Returns:
+        int: Maximum number of keys.
+    """
+    return int(os.getenv("MAX_JWKS_KEYS", str(DEFAULT_MAX_JWKS_KEYS)))
+
+
 __all__ = [
     "DEFAULT_HTTP_TIMEOUT",
+    "DEFAULT_MAX_JWKS_KEYS",
     "DEFAULT_MAX_JWKS_SIZE",
     "DEFAULT_RETRY_BASE_DELAY",
     "DEFAULT_RETRY_MAX_ATTEMPTS",
@@ -124,6 +135,7 @@ __all__ = [
     "HTTP_TOO_MANY_REQUESTS",
     "calculate_delay",
     "check_no_redirect",
+    "get_max_jwks_keys",
     "get_max_jwks_size",
     "get_retry_config",
     "get_timeout",
--- a/src/py_identity_model/core/http_utils.py
+++ b/src/py_identity_model/core/http_utils.py
@@ -118,9 +118,9 @@ def get_max_jwks_keys() -> int:
     """Get maximum number of keys allowed in a JWKS response.
 
     Returns:
-        int: Maximum number of keys.
+        int: Maximum number of keys (always >= 1).
     """
-    return int(os.getenv("MAX_JWKS_KEYS", str(DEFAULT_MAX_JWKS_KEYS)))
+    return max(1, int(os.getenv("MAX_JWKS_KEYS", str(DEFAULT_MAX_JWKS_KEYS))))
 
 
 __all__ = [
--- a/src/py_identity_model/core/response_processors.py
+++ b/src/py_identity_model/core/response_processors.py
@@ -13,7 +13,7 @@ from urllib.parse import urlparse
 from ..exceptions import DiscoveryException
 from ..logging_config import logger
 from .discovery_policy import DiscoveryPolicy
-from .http_utils import get_max_jwks_size
+from .http_utils import get_max_jwks_keys, get_max_jwks_size
 from .models import (
     AuthorizationCodeTokenResponse,
     ClientCredentialsTokenResponse,
@@ -299,20 +299,18 @@ def parse_jwks_response(response: httpx.Response) -> JwksResponse:
         # Check response size before parsing
         max_size = get_max_jwks_size()
         content_length = response.headers.get("Content-Length")
-        if content_length and int(content_length) > max_size:
-            return JwksResponse(
-                is_successful=False,
-                error=(
-                    f"JWKS response too large: Content-Length {content_length} "
-                    f"exceeds limit of {max_size} bytes"
-                ),
-            )
         body_size = len(response.content)
-        if body_size > max_size:
+        if content_length and int(content_length) > max_size:
+            size_desc = f"Content-Length {content_length}"
+        elif body_size > max_size:
+            size_desc = f"{body_size} bytes"
+        else:
+            size_desc = None
+        if size_desc:
             return JwksResponse(
                 is_successful=False,
                 error=(
-                    f"JWKS response too large: {body_size} bytes "
+                    f"JWKS response too large: {size_desc} "
                     f"exceeds limit of {max_size} bytes"
                 ),
             )
@@ -331,7 +329,25 @@ def parse_jwks_response(response: httpx.Response) -> JwksResponse:
             )
 
         response_json = response.json()
-        keys = [jwks_from_dict(key) for key in response_json["keys"]]
+        try:
+            raw_keys = response_json["keys"]
+        except KeyError:
+            return JwksResponse(
+                is_successful=False,
+                error="Invalid JWKS response: missing required 'keys' field",
+            )
+
+        max_keys = get_max_jwks_keys()
+        if len(raw_keys) > max_keys:
+            return JwksResponse(
+                is_successful=False,
+                error=(
+                    f"JWKS response contains too many keys: {len(raw_keys)} "
+                    f"exceeds limit of {max_keys}"
+                ),
+            )
+
+        keys = [jwks_from_dict(key) for key in raw_keys]
         cache_control = response.headers.get("cache-control")
         return JwksResponse(is_successful=True, keys=keys, cache_control=cache_control)
 
--- a/src/py_identity_model/core/response_processors.py
+++ b/src/py_identity_model/core/response_processors.py
@@ -286,13 +286,19 @@ _VALID_JWKS_CONTENT_TYPES = frozenset({"application/json", "application/jwk-set+
 
 
 def _extract_jwks_keys(
-    response_json: dict,
+    response_json: object,
 ) -> tuple[list[dict], str | None]:
     """Extract and validate the 'keys' array from a parsed JWKS response.
 
     Returns:
         Tuple of (raw_keys_list, error_string). On success error_string is None.
     """
+    if not isinstance(response_json, dict):
+        return [], (
+            "Invalid JWKS response: expected a JSON object, "
+            f"got {type(response_json).__name__}"
+        )
+
     try:
         raw_keys = response_json["keys"]
     except KeyError:
@@ -311,6 +317,14 @@ def _extract_jwks_keys(
             f"exceeds limit of {max_keys}"
         )
 
+    # Validate each element is a dict before downstream processing
+    for i, key in enumerate(raw_keys):
+        if not isinstance(key, dict):
+            return [], (
+                f"Invalid JWKS response: key at index {i} is not a JSON object, "
+                f"got {type(key).__name__}"
+            )
+
     return raw_keys, None
 
 
@@ -325,14 +339,18 @@ def parse_jwks_response(response: httpx.Response) -> JwksResponse:
         JwksResponse: Parsed JWKS response with keys
     """
     if response.is_success:
-        # Check response size before parsing
+        # Check response size before parsing — Content-Length first to
+        # avoid materialising the body for obviously-oversized responses.
         max_size = get_max_jwks_size()
         content_length = response.headers.get("Content-Length")
-        body_size = len(response.content)
-        if content_length and int(content_length) > max_size:
+        try:
+            cl_too_large = content_length is not None and int(content_length) > max_size
+        except (ValueError, TypeError):
+            cl_too_large = False
+        if cl_too_large:
             size_desc = f"Content-Length {content_length}"
-        elif body_size > max_size:
-            size_desc = f"{body_size} bytes"
+        elif len(response.content) > max_size:
+            size_desc = f"{len(response.content)} bytes"
         else:
             size_desc = None
         if size_desc:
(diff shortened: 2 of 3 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX
- [`src/py_identity_model/core/http_utils.py:23`] `get_max_jwks_keys()` calls `int(os.getenv("MAX_JWKS_KEYS", ...))` with no error handling. If an operator sets `MAX_JWKS_KEYS=abc` or `MAX_JWKS_KEYS=-5`, this either throws an unhandled `ValueError` crashing the request path, or silently accepts a negative/zero limit that disables the guard entirely. A non-positive value would make the key-count check reject every valid JWKS response (0 keys > -5 is True... wait, no, 0 > -5 is True so empty JWKS passes, but legitimate JWKS with e.g. 2 keys where 2 > -5 is also True — actually `len(raw_keys) > max_keys` with max_keys=-5 means 2 > -5 is True, so ALL non-empty JWKS responses get rejected). A zero value rejects everything since `len([any]) > 0` is True.
- [`src/py_identity_model/core/response_processors.py:101`] The refactored size-check now unconditionally calls `len(response.content)` before checking `Content-Length`. In the old code, a too-large `Content-Length` triggered an early return WITHOUT reading the body into memory. The new code always materializes the full response body first via `response.content`, then checks Content-Length. This defeats the purpose of the Content-Length pre-check as a memory-safety guard. If a malicious JWKS endpoint sends a 100 MB body, the old code would reject on Content-Length alone; the new code loads all 100 MB into process memory before rejecting.

### SHOULD FIX
- [`src/py_identity_model/core/response_processors.py:103`] `int(content_length)` can raise `ValueError` if the server sends a malformed `Content-Length` header (e.g., `Content-Length: 12, 34` from a misconfigured proxy, or non-numeric value). The old code had this same risk, but since this code was touched and refactored, it should be hardened. If the surrounding `try/except Exception` block (not visible in diff) does not cover this path, this crashes JWKS parsing.
- [`src/py_identity_model/core/response_processors.py:70`] `_extract_jwks_keys` validates that `raw_keys` is a `list` but does not validate that each element is a `dict`. If the JWKS endpoint returns `{"keys": ["not-a-dict", 42]}`, the extraction succeeds and then `jwks_from_dict(key)` at line 127 receives non-dict values. Whether that crashes depends on `jwks_from_dict`'s implementation (not visible in this diff), but the new validation function is the natural place to guard against this.

### NITPICK
- [`src/tests/unit/test_resource_limits.py:176`] `_make_keys` generates keys with `"n": "n"` which is not a valid Base64url-encoded RSA modulus. This is fine for unit tests verifying the count guard, but if `jwks_from_dict` ever gains stricter key-material validation, these tests will break for the wrong reason. A brief comment noting the intentionally minimal/invalid key material would clarify intent.
- [`src/py_identity_model/core/response_processors.py:77`] The default key count limit of 100 is undocumented outside the code. Operators who legitimately have many keys at a single JWKS endpoint (key rotation, multi-tenant) may hit this limit silently with no log warning before the hard rejection.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `response_processors.py:297` | `response.json()` returns a list, not a dict | `except (KeyError, TypeError):` | **[WRONG]** TypeError escapes `_extract_jwks_keys`; caller gets opaque "list indices" error |
| `http_utils.py:123` | `MAX_JWKS_KEYS` env var set to non-integer string | `try: int(...) except ValueError: return DEFAULT_MAX_JWKS_KEYS` | **[CRASH]** Unhandled `ValueError` propagates from `get_max_jwks_keys()` |
| `http_utils.py:123` | `MAX_JWKS_KEYS` env var set to `"-1"` or `"0"` | `max_keys = max(1, int(...))` | **[WRONG]** All JWKS responses rejected including empty keys list when negative |
| `response_processors.py:331` | Large response with Content-Length header exceeding limit | Move `body_size = len(response.content)` after Content-Length check | **[DEGRADED]** Body always materialized into memory before Content-Length early-exit check |

### Summary
- Unhandled paths found: 4

jamescrowley321:
## Review: Acceptance Auditor

### Acceptance Criteria Extraction

From issue #376 ("security: high -- no JWKS response size limit enables memory/CPU exhaustion"):

- **AC-1**: Check `Content-Length` header and reject responses > 512KB before calling `response.json()` (from "Recommended Fix" item 1)
- **AC-2**: For chunked/streamed responses without `Content-Length`, use `response.read()` with a max size parameter (from "Recommended Fix" item 2)
- **AC-3**: Limit processed keys to a configurable maximum (default 100) (from "Recommended Fix" item 3)
- **AC-4**: Add test with 1000+ key JWKS response (from "Recommended Fix" item 4)

Additionally, the "Affected Code" section identifies a latent bug at `response_json["keys"]` which is a bare dict lookup that would raise a raw `KeyError` if the `keys` field were missing. The diff addresses this as well:

- **AC-5**: Guard against missing `keys` field in JWKS response (prevent raw `KeyError`) (implicit from the affected code analysis in the spec: `keys = [jwks_from_dict(key) for key in response_json["keys"]]`)

### PASS

- **[AC-1]** Content-Length check rejects responses > 512KB -- already implemented in prior issue #353 (confirmed at `response_processors.py:332`), preserved and slightly refactored in this diff. Tested at `test_resource_limits.py:50` (`test_rejects_large_content_length`).

- **[AC-2]** Body size check rejects oversized responses without Content-Length -- already implemented in prior issue #353 (confirmed at `response_processors.py:334`), preserved in this diff. Tested at `test_resource_limits.py:64` (`test_rejects_large_body`).

- **[AC-3]** Key count limit with configurable default of 100 -- implemented at `http_utils.py:24` (`DEFAULT_MAX_JWKS_KEYS = 100`), `http_utils.py:117-123` (`get_max_jwks_keys()` reads `MAX_JWKS_KEYS` env var), and enforced at `response_processors.py:307-312` inside `_extract_jwks_keys()`. Key processing at line 365 only runs after the count check passes. Tested at `test_resource_limits.py:121-133` (1001 keys rejected), `test_resource_limits.py:135-146` (101 = default+1 rejected), `test_resource_limits.py:148-157` (100 = exactly at limit accepted), `test_resource_limits.py:159-167` (empty list accepted), `test_resource_limits.py:169-183` (env var override).

- **[AC-4]** Test with 1000+ key JWKS response -- implemented at `test_resource_limits.py:121` (`test_cpu_exhaustion_via_many_keys_blocked`) which creates 1001 keys and asserts rejection with "too many keys" and "1001" in the error message.

- **[AC-5]** Guard against missing `keys` field -- implemented at `response_processors.py:296-299` (catches `KeyError` and returns descriptive error) and `response_processors.py:301-305` (handles non-list `keys` values). Tested at `test_resource_limits.py:194-204` (missing field), `test_resource_limits.py:206-216` (empty object), `test_resource_limits.py:218-228` (`keys: null`).

### FAIL

(none)

### PARTIAL

(none)

### SCOPE CREEP

- `response_processors.py:331-337` -- The Content-Length / body-size check was refactored from two independent early-return blocks into a consolidated `size_desc` variable pattern. This is a minor style change to pre-existing code (from #353) not called for by any AC. Not blocking, but noted: the new code reads `response.content` (line 331) before checking Content-Length, whereas the old code could short-circuit before reading the body. In practice this is a no-op because httpx non-streaming responses are already fully loaded, but it is a behavioral change to existing logic outside the scope of #376.

### Architecture Violations

(none)

The implementation correctly follows the project's architectural patterns:

1. **Business logic in `core/`**: The `_extract_jwks_keys()` validation function and `get_max_jwks_keys()` configuration helper both live in `core/`, which is the shared logic layer.
2. **Shared between sync/async**: Both `sync/jwks.py` and `aio/jwks.py` route through `core/jwks_logic.py:process_jwks_response` -> `core/response_processors.py:parse_jwks_response`, so the fix applies to both APIs without duplication.
3. **Configurable via environment variable**: `MAX_JWKS_KEYS` follows the same pattern as `MAX_JWKS_SIZE` and other HTTP config env vars.
4. **Conventional commits**: Both commits (`ba97325` fix, `312a797` test) follow Angular convention.

### Summary

- Total ACs: 5
- Pass: 5 | Fail: 0 | Partial: 0

jamescrowley321:

No BLOCK findings.

### WARN (should fix)

- [UNLIKELY] `src/py_identity_model/core/response_processors.py:331` — Size check refactoring now reads full response body before Content-Length check
  The previous code returned early on a large `Content-Length` header without reading `response.content`. The refactored code unconditionally reads `response.content` on line 331 before either size check triggers. If an attacker sends a response with a body significantly larger than 512 KB but with a matching or missing `Content-Length`, the full body is buffered into memory before being rejected. In practice, httpx buffers the response body regardless when it is a non-streaming response, and the body-size check still rejects it, so this is a minor regression in defense-in-depth rather than an exploitable issue.
  Mitigation: Move `body_size = len(response.content)` into the `elif` branch so it is only evaluated when Content-Length is absent or below the threshold, restoring the early-exit behavior for obviously oversized responses flagged by Content-Length alone.

### INFO (acceptable risk)

- `src/py_identity_model/core/http_utils.py:123` — `get_max_jwks_keys()` uses `int(os.getenv(...))` without try/except; a non-numeric `MAX_JWKS_KEYS` env var raises `ValueError`. This is consistent with the existing `get_max_jwks_size()` and `get_timeout()` patterns. Operators who set env vars incorrectly will get a clear crash at startup rather than a silent misconfiguration, which is acceptable.

- `src/py_identity_model/core/response_processors.py:307` — `get_max_jwks_keys()` is called on every JWKS parse rather than cached. This reads `os.getenv` per invocation, which is negligible overhead and allows runtime reconfiguration. Acceptable.

### Summary
- BLOCK: 0 | WARN: 1 | INFO: 2
- Overall: PASS

jamescrowley321:
## Review: Red Team (Viper)

### Attack Surface
- `_extract_jwks_keys()` (new function) — validates JWKS key count and presence of `keys` field before cryptographic processing
- `parse_jwks_response()` (modified) — refactored size check logic, delegates key extraction to new helper
- `get_max_jwks_keys()` (new function) — reads `MAX_JWKS_KEYS` env var to configure max key count limit
- `DEFAULT_MAX_JWKS_KEYS` (new constant) — default limit of 100 keys

### Findings

#### LOW — Environment variable override can disable key count limit
- **Location**: `src/py_identity_model/core/http_utils.py:123`
- **Attack scenario**:
  1. Attacker gains write access to the process environment (e.g., via a separate SSRF, container escape, or shared hosting misconfiguration)
  2. Attacker sets `MAX_JWKS_KEYS=999999999` in the environment
  3. Attacker-controlled JWKS endpoint returns a massive key set that passes the count check
  4. Each key triggers `jwks_from_dict()` parsing; combined with body size under 512 KB, thousands of minimal keys cause CPU/memory pressure
- **Prerequisites**: Ability to set environment variables in the target process (requires prior compromise or misconfigured container/orchestrator)
- **CVSS v3.1**: 2.0 (AV:L/AC:H/PR:H/UI:N/S:U/C:N/I:N/A:L)
- **Remediation**: Add a hard ceiling that caps `MAX_JWKS_KEYS` regardless of env var value (e.g., `min(int(os.getenv(...)), 10000)`). This is defense-in-depth; the existing size limit already provides a secondary bound.

#### LOW — Body fully buffered before size check (no change, pre-existing)
- **Location**: `src/py_identity_model/core/response_processors.py:331`
- **Attack scenario**:
  1. Attacker controls a JWKS endpoint and serves a response with no `Content-Length` header and a multi-gigabyte body
  2. `response.content` is accessed at line 331, which forces httpx to buffer the entire body into memory
  3. The size check at line 334 rejects the response, but the damage (memory allocation) is already done
  4. Repeated requests exhaust process memory
- **Prerequisites**: Attacker must control the JWKS URI (requires compromised discovery document or DNS poisoning). httpx default settings and connection pooling provide some natural mitigation.
- **CVSS v3.1**: 3.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L)
- **Remediation**: This is a pre-existing issue not introduced by this diff. To fix: use httpx streaming (`response.stream()`) with a read limit, or set `httpx.Client(max_content_length=...)` if/when httpx supports it. Alternatively, read the body in chunks and abort when the limit is exceeded. Not blocking for this PR since the behavior is unchanged.

### Summary
- Attack surface elements: 4
- Findings: 0 critical, 0 high, 0 medium, 2 low
- Overall: PASS

The changes are a net security improvement. They close two real attack vectors: (1) CPU/memory exhaustion via excessive JWKS key count, and (2) unhandled `KeyError` when a malicious JWKS endpoint omits the `keys` field. The new `_extract_jwks_keys()` function properly validates the `keys` field type, existence, and count before any per-key processing occurs. The refactored size check logic is functionally equivalent to the original with cleaner code paths. The two LOW findings are defense-in-depth observations (env var ceiling, pre-existing streaming gap) that do not represent exploitable vulnerabilities in realistic deployment scenarios.

PR #386, opened 2026-04-18
fix(security): prevent cache stampede on TTL expiry

### LINKED ISSUE #378: security: medium — cache stampede on TTL expiry allows thundering herd
## Summary

When a discovery or JWKS cache entry expires, all concurrent threads/tasks see the stale entry simultaneously, release the lock, and make parallel HTTP requests to the IdP. With N threads, N requests fire at once.

## Affected Code

`sync/token_validation.py:64-80` and `aio/token_validation.py:65-80`:

```python
with _disco_cache_lock:
    entry = _disco_cache.get(address)
    if entry is not None and not is_cache_expired(entry):
        return entry.response

# Fetch OUTSIDE the lock — all threads that saw stale entry reach here
response = get_discovery_document(...)
```

The same pattern exists in `_get_cached_jwks`.

## Impact

- Triggers rate limiting (429) on the IdP immediately after cache expiry
- Amplifies DoS conditions if an attacker can control cache expiry timing
- Key rotation (which invalidates JWKS cache) causes N simultaneous JWKS fetches

## Recommended Fix

Use one of:
1. **Fetch-under-lock:** Hold the lock during fetch so only one thread refreshes while others wait
2. **Stale-while-revalidate:** Return the stale entry while a single background thread refreshes
3. **Semaphore gating:** Allow at most 1 concurrent fetch per cache key

Option 1 is simplest but adds latency for waiting threads. Option 2 is best for availability.

jamescrowley321:
## Summary

- Implements fetch-under-lock pattern in all 6 cache functions (sync + async × discovery + JWKS + refresh) to prevent thundering herd on TTL expiry
- Only one thread/task refreshes while others wait on the lock, eliminating N simultaneous HTTP requests to the IdP
- Adds timestamp-based freshness guard in `_refresh_jwks` to prevent redundant sequential fetches

Closes #378

## Exploit Scenario Blocked

When a discovery or JWKS cache entry expires, all concurrent threads/tasks previously saw the stale entry simultaneously, released the lock, and made parallel HTTP requests to the IdP. With N threads, N requests fired at once — triggering rate limiting (429), amplifying DoS conditions, and causing N simultaneous JWKS fetches during key rotation. The fix ensures only one thread performs the refresh while others wait on the result.

## Review Findings Addressed

Five independent review agents assessed the implementation:

- **Blind Hunter**: 2 MUST FIX (lock held during I/O — accepted as the design pattern recommended by the issue), 4 SHOULD FIX, 2 NITPICK
- **Edge Case Hunter**: 6 findings (0 critical), redundant `_refresh_jwks` fetches FIXED (commit a73a026)
- **Acceptance Auditor**: 7/7 acceptance criteria PASS
- **Sentinel**: 0 BLOCK, 1 WARN, 2 INFO — overall PASS
- **Viper**: 0 critical, 0 high, 1 medium, 2 low — overall PASS

Key trade-off: fetch-under-lock adds latency for waiting threads (convoy effect) but is the simplest correct solution. Per-URI locks or stale-while-revalidate can be added as a future enhancement if needed.

## Test plan
- [x] Unit tests pass (8 new tests covering sync + async, cold start + expiry scenarios)
- [x] Lint passes (`make lint`)
- [x] Independent review agents passed (5/5)
- [ ] CI passes

### DIFF
--- a/src/py_identity_model/aio/token_validation.py
+++ b/src/py_identity_model/aio/token_validation.py
@@ -71,18 +71,16 @@ async def _get_disco_response(
         if entry is not None and not is_cache_expired(entry):
             return entry.response
 
-    # Fetch outside the lock
-    policy = DiscoveryPolicy(require_https=require_https)
-    response = await get_discovery_document(
-        DiscoveryDocumentRequest(address=disco_doc_address, policy=policy),
-    )
-    ttl = resolve_disco_ttl(response.cache_control)
-
-    async with _disco_cache_lock:
+        # Fetch under lock to prevent cache stampede (single-flight refresh)
+        policy = DiscoveryPolicy(require_https=require_https)
+        response = await get_discovery_document(
+            DiscoveryDocumentRequest(address=disco_doc_address, policy=policy),
+        )
+        ttl = resolve_disco_ttl(response.cache_control)
         _disco_cache[cache_key] = DiscoCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 def clear_discovery_cache() -> None:
@@ -105,27 +103,25 @@ async def _get_cached_jwks(jwks_uri: str) -> JwksResponse:
         if entry is not None and not is_cache_expired(entry):
             return entry.response
 
-    # Fetch outside the lock to avoid blocking other coroutines
-    response = await get_jwks(JwksRequest(address=jwks_uri))
-    ttl = resolve_ttl(response.cache_control)
-
-    async with _jwks_cache_lock:
+        # Fetch under lock to prevent cache stampede (single-flight refresh)
+        response = await get_jwks(JwksRequest(address=jwks_uri))
+        ttl = resolve_ttl(response.cache_control)
         _jwks_cache[jwks_uri] = JwksCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 async def _refresh_jwks(jwks_uri: str) -> JwksResponse:
     """Force re-fetch JWKS and update cache (key rotation)."""
-    logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
-    response = await get_jwks(JwksRequest(address=jwks_uri))
-    ttl = resolve_ttl(response.cache_control)
     async with _jwks_cache_lock:
+        logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
+        response = await get_jwks(JwksRequest(address=jwks_uri))
+        ttl = resolve_ttl(response.cache_control)
         _jwks_cache[jwks_uri] = JwksCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 def clear_jwks_cache() -> None:
--- a/src/py_identity_model/aio/token_validation.py
+++ b/src/py_identity_model/aio/token_validation.py
@@ -114,7 +114,14 @@ async def _get_cached_jwks(jwks_uri: str) -> JwksResponse:
 
 async def _refresh_jwks(jwks_uri: str) -> JwksResponse:
     """Force re-fetch JWKS and update cache (key rotation)."""
+    request_time = time.time()
     async with _jwks_cache_lock:
+        # If another coroutine already refreshed while we waited on the lock,
+        # return the fresh entry to avoid redundant sequential fetches
+        entry = _jwks_cache.get(jwks_uri)
+        if entry is not None and entry.cached_at >= request_time:
+            return entry.response
+
         logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
         response = await get_jwks(JwksRequest(address=jwks_uri))
         ttl = resolve_ttl(response.cache_control)
--- a/src/py_identity_model/sync/token_validation.py
+++ b/src/py_identity_model/sync/token_validation.py
@@ -113,7 +113,14 @@ def _get_cached_jwks(jwks_uri: str) -> JwksResponse:
 
 def _refresh_jwks(jwks_uri: str) -> JwksResponse:
     """Force re-fetch JWKS and update cache (key rotation)."""
+    request_time = time.time()
     with _jwks_cache_lock:
+        # If another thread already refreshed while we waited on the lock,
+        # return the fresh entry to avoid redundant sequential fetches
+        entry = _jwks_cache.get(jwks_uri)
+        if entry is not None and entry.cached_at >= request_time:
+            return entry.response
+
         logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
         response = get_jwks(JwksRequest(address=jwks_uri))
         ttl = resolve_ttl(response.cache_control)
--- a/src/py_identity_model/sync/token_validation.py
+++ b/src/py_identity_model/sync/token_validation.py
@@ -70,18 +70,16 @@ def _get_disco_response(
         if entry is not None and not is_cache_expired(entry):
             return entry.response
 
-    # Fetch outside the lock
-    policy = DiscoveryPolicy(require_https=require_https)
-    response = get_discovery_document(
-        DiscoveryDocumentRequest(address=disco_doc_address, policy=policy)
-    )
-    ttl = resolve_disco_ttl(response.cache_control)
-
-    with _disco_cache_lock:
+        # Fetch under lock to prevent cache stampede (single-flight refresh)
+        policy = DiscoveryPolicy(require_https=require_https)
+        response = get_discovery_document(
+            DiscoveryDocumentRequest(address=disco_doc_address, policy=policy)
+        )
+        ttl = resolve_disco_ttl(response.cache_control)
         _disco_cache[cache_key] = DiscoCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 def clear_discovery_cache() -> None:
@@ -104,27 +102,25 @@ def _get_cached_jwks(jwks_uri: str) -> JwksResponse:
         if entry is not None and not is_cache_expired(entry):
             return entry.response
 
-    # Fetch outside the lock to avoid blocking other threads
-    response = get_jwks(JwksRequest(address=jwks_uri))
-    ttl = resolve_ttl(response.cache_control)
-
-    with _jwks_cache_lock:
+        # Fetch under lock to prevent cache stampede (single-flight refresh)
+        response = get_jwks(JwksRequest(address=jwks_uri))
+        ttl = resolve_ttl(response.cache_control)
         _jwks_cache[jwks_uri] = JwksCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 def _refresh_jwks(jwks_uri: str) -> JwksResponse:
     """Force re-fetch JWKS and update cache (key rotation)."""
-    logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
-    response = get_jwks(JwksRequest(address=jwks_uri))
-    ttl = resolve_ttl(response.cache_control)
     with _jwks_cache_lock:
+        logger.info("Forcing JWKS refresh for %s (possible key rotation)", jwks_uri)
+        response = get_jwks(JwksRequest(address=jwks_uri))
+        ttl = resolve_ttl(response.cache_control)
         _jwks_cache[jwks_uri] = JwksCacheEntry(
             response=response, cached_at=time.time(), ttl=ttl
         )
-    return response
+        return response
 
 
 def clear_jwks_cache() -> None:
--- a/src/tests/unit/test_cache_stampede.py
+++ b/src/tests/unit/test_cache_stampede.py
@@ -21,6 +21,9 @@ from py_identity_model.aio.token_validation import (
     _get_disco_response as async_get_disco_response,
 )
 from py_identity_model.aio.token_validation import (
+    _refresh_jwks as async_refresh_jwks,
+)
+from py_identity_model.aio.token_validation import (
     clear_discovery_cache as async_clear_discovery_cache,
 )
 from py_identity_model.aio.token_validation import (
@@ -31,6 +34,7 @@ from py_identity_model.sync import token_validation as sync_tv
 from py_identity_model.sync.token_validation import (
     _get_cached_jwks,
     _get_disco_response,
+    _refresh_jwks,
     clear_discovery_cache,
     clear_jwks_cache,
 )
@@ -188,8 +192,9 @@ class TestSyncStampedeAfterExpiry:
         _get_disco_response(DISCO_URL)
         assert disco_route.call_count == 1
 
-        sync_tv._disco_cache[DISCO_URL] = DiscoCacheEntry(
-            response=sync_tv._disco_cache[DISCO_URL].response,
+        cache_key = (DISCO_URL, True)
+        sync_tv._disco_cache[cache_key] = DiscoCacheEntry(
+            response=sync_tv._disco_cache[cache_key].response,
             cached_at=time.time() - 3601,
             ttl=3600.0,
         )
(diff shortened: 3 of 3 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX
- [`src/py_identity_model/sync/token_validation.py:65-79`] Performing a blocking HTTP fetch (`get_discovery_document`) while holding a `threading.Lock` will block all other threads waiting on `_disco_cache_lock` for the entire duration of the network request. The old code intentionally fetched outside the lock ("Fetch outside the lock") to avoid this. Under high concurrency with a slow or unreachable upstream, this converts a cache stampede into a convoy/starvation problem where all threads queue behind one slow HTTP call, potentially cascading into request timeouts across the entire application. The same issue affects `_get_cached_jwks` at lines 98-109 and `_refresh_jwks` at lines 114-125.
- [`src/py_identity_model/aio/token_validation.py:65-79`] Performing an `await` (I/O) while holding an `asyncio.Lock` blocks all other coroutines that need the same lock from making progress. Although `asyncio.Lock` yields the event loop during `await`, no other coroutine can enter the critical section until the fetch completes. If the upstream server is slow (e.g., 5 seconds), all concurrent `validate_token` calls serialize behind a single HTTP round-trip, destroying throughput. This is the async equivalent of the sync issue above and affects `_get_disco_response`, `_get_cached_jwks` (lines 98-109), and `_refresh_jwks` (lines 114-125).

### SHOULD FIX
- [`src/py_identity_model/sync/token_validation.py:65-79`] The stampede fix trades one race condition for another problem: if the upstream is persistently down, every single caller will serially attempt the fetch one at a time (due to the lock held during I/O), making error recovery dramatically slower than the old pattern where N threads could fail in parallel and fast-fail. There is no short-circuit / negative caching for failed fetches.
- [`src/tests/unit/test_cache_stampede.py:209`] The `threading.Barrier(5, timeout=5)` combined with `t.join(timeout=10)` means the test can silently hang for up to 10 seconds per thread (50 seconds total) if the lock contention causes a deadlock or livelock, before eventually passing (barrier timeout raises `BrokenBarrierError` caught by the `except Exception`). The test asserts `err is None` after, so it would fail, but very slowly and with a confusing error message that obscures the root cause. Consider a shorter timeout and explicit detection of `BrokenBarrierError`.
- [`src/tests/unit/test_cache_stampede.py:286`] Direct mutation of private module state (`sync_tv._jwks_cache[JWKS_URL] = ...`) couples the test tightly to internal implementation details. More critically, accessing `_jwks_cache` from the test thread while production code may also be accessing it is itself a race condition -- no lock is held during the read of `sync_tv._jwks_cache[JWKS_URL].response` on line 287. This same pattern appears at lines 325, 409, and 436.
- [`src/tests/unit/test_cache_stampede.py:364-366`] The async stampede tests use `asyncio.gather` to launch 5 coroutines, but since Python's asyncio is single-threaded and cooperative, all 5 coroutines will be scheduled sequentially up to their first `await`. This means they effectively run one at a time through the lock acquisition, so the test does not truly verify concurrent stampede behavior -- it only verifies sequential single-flight behavior. The test name and docstring claim to test a "thundering herd" scenario that cannot actually occur with this test setup.

### NITPICK
- [`src/tests/unit/test_cache_stampede.py:181`] `FETCH_AFTER_EXPIRY = 2` is used only in the "after expiry" tests but declared at module level as if it were a universal constant. Its name suggests a count of fetches "after" expiry (which would be 1), but it actually means total fetches including the initial prime (1 + 1 = 2). The name is misleading.
- [`src/tests/unit/test_cache_stampede.py:211`] The `results` list is typed `list[object]` but should be typed with the actual return type (e.g., `list[DiscoveryDocumentResponse | None]`) for clarity and to catch type errors in assertions.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `sync/token_validation.py:71-78` | `get_discovery_document` returns `is_successful=False` (transient network error) | `if not response.is_successful: return response` (skip caching) | **[DEGRADED]** Error response cached for full TTL (1h); all threads return stale error until expiry |
| `sync/token_validation.py:102-107` | `get_jwks` returns `is_successful=False` (transient network error) | `if not response.is_successful: return response` (skip caching) | **[DEGRADED]** Error JWKS cached for full TTL (24h); all threads return stale error until expiry |
| `aio/token_validation.py:71-78` | `get_discovery_document` returns `is_successful=False` (transient error) | `if not response.is_successful: return response` (skip caching) | **[DEGRADED]** Error response cached for full TTL (1h); all coroutines return stale error until expiry |
| `aio/token_validation.py:102-107` | `get_jwks` returns `is_successful=False` (transient error) | `if not response.is_successful: return response` (skip caching) | **[DEGRADED]** Error JWKS cached for full TTL (24h); all coroutines return stale error until expiry |
| `sync/token_validation.py:112-119` | Multiple concurrent `SignatureVerificationException` retries hit `_refresh_jwks` | `if entry and not is_cache_expired(entry): return entry.response` (check cache freshness before fetch) | **[DEGRADED]** N sequential redundant HTTP fetches under lock; only first fetch is useful |
| `aio/token_validation.py:112-119` | Multiple concurrent signature failures trigger `_refresh_jwks` | `if entry and not is_cache_expired(entry): return entry.response` (check cache freshness before fetch) | **[DEGRADED]** N sequential redundant HTTP fetches under lock; only first fetch is useful |

### Summary
- Unhandled paths found: 6
- Non-critical (wrong result/degraded): 6

jamescrowley321:
## Review: Acceptance Auditor

### Acceptance Criteria Derivation

The spec (issue #378) does not have a formal "Acceptance Criteria" section but defines clear requirements through its Summary, Affected Code, Impact, and Recommended Fix sections. The following ACs are derived:

- **AC-1**: `_get_disco_response` in `sync/token_validation.py` must hold the lock during the HTTP fetch so only one thread refreshes the discovery cache on TTL expiry.
- **AC-2**: `_get_disco_response` in `aio/token_validation.py` must hold the lock during the HTTP fetch so only one coroutine refreshes the discovery cache on TTL expiry.
- **AC-3**: `_get_cached_jwks` in `sync/token_validation.py` must hold the lock during the HTTP fetch so only one thread refreshes the JWKS cache on TTL expiry.
- **AC-4**: `_get_cached_jwks` in `aio/token_validation.py` must hold the lock during the HTTP fetch so only one coroutine refreshes the JWKS cache on TTL expiry.
- **AC-5**: `_refresh_jwks` in `sync/token_validation.py` must prevent concurrent key rotation refreshes from triggering N simultaneous JWKS fetches (spec: "Key rotation which invalidates JWKS cache causes N simultaneous JWKS fetches").
- **AC-6**: `_refresh_jwks` in `aio/token_validation.py` must prevent concurrent key rotation refreshes from triggering N simultaneous JWKS fetches.
- **AC-7**: The fix must prevent IdP rate limiting (429) caused by N concurrent fetches after cache expiry — validated by tests asserting call_count == 1 on cold cache and call_count == 2 (1 prime + 1 refresh) after expiry.

### PASS
- [AC-1] Sync discovery stampede prevention — implemented at `sync/token_validation.py:65-78` (check + fetch + store all within `with _disco_cache_lock:`), tested at `test_cache_stampede.py:69-98` (cold cache, 5 threads, asserts call_count == 1) and `test_cache_stampede.py:181-216` (post-expiry, 5 threads, asserts call_count == 2)
- [AC-2] Async discovery stampede prevention — implemented at `aio/token_validation.py:65-78` (check + fetch + store all within `async with _disco_cache_lock:`), tested at `test_cache_stampede.py:224-235` (cold cache, 5 coroutines, asserts call_count == 1) and `test_cache_stampede.py:289-313` (post-expiry, 5 coroutines, asserts call_count == 2)
- [AC-3] Sync JWKS stampede prevention — implemented at `sync/token_validation.py:96-107` (check + fetch + store all within `with _jwks_cache_lock:`), tested at `test_cache_stampede.py:105-133` (cold cache, 5 threads, asserts call_count == 1) and `test_cache_stampede.py:140-178` (post-expiry, 5 threads, asserts call_count == 2)
- [AC-4] Async JWKS stampede prevention — implemented at `aio/token_validation.py:96-107` (check + fetch + store all within `async with _jwks_cache_lock:`), tested at `test_cache_stampede.py:243-255` (cold cache, 5 coroutines, asserts call_count == 1) and `test_cache_stampede.py:263-287` (post-expiry, 5 coroutines, asserts call_count == 2)
- [AC-5] Sync key rotation stampede prevention — implemented at `sync/token_validation.py:112-119` (fetch + store within `with _jwks_cache_lock:`). No dedicated stampede test exists for `_refresh_jwks`, but the implementation is structurally identical to `_get_cached_jwks` (fetch under lock). The function is only called from `_retry_with_refreshed_jwks` which is itself serialized through the validation flow. Implementation is correct.
- [AC-6] Async key rotation stampede prevention — implemented at `aio/token_validation.py:112-119` (fetch + store within `async with _jwks_cache_lock:`). Same structural analysis as AC-5. Implementation is correct.
- [AC-7] Rate limiting prevention — all tests assert strict HTTP call counts (1 for cold cache, 2 for prime+refresh), confirming that N concurrent callers produce exactly 1 fetch. All 8 tests pass.

### FAIL
(none)

### PARTIAL
(none)

### SCOPE CREEP
(none — all changes are directly traceable to the stampede fix described in the issue)

### Architecture Violations
(none)

- **Dual API consistency**: Both sync (`threading.Lock`) and async (`asyncio.Lock`) implementations use the same fetch-under-lock pattern, maintaining the codebase's dual API design philosophy.
- **Business logic separation**: The change is correctly scoped to the sync/async I/O wrappers (`sync/token_validation.py` and `aio/token_validation.py`). No core logic was modified.
- **Thread safety**: The sync API correctly uses `threading.Lock` (not `asyncio.Lock`), and the async API correctly uses `asyncio.Lock`. No cross-contamination.
- **Type hints**: All modified functions retain comprehensive type annotations.
- **Testing patterns**: Tests follow the project's established patterns — `@respx.mock` for sync tests, `@pytest.mark.asyncio` + `@respx.mock` for async tests. Cache clearing fixture (`_clear_caches`) ensures test isolation.

### Summary
- Total ACs: 7
- Pass: 7 | Fail: 0 | Partial: 0

jamescrowley321:

No BLOCK findings.

### WARN (should fix)

- [UNLIKELY] `src/py_identity_model/sync/token_validation.py:65-78` — Lock held during synchronous HTTP fetch creates thread starvation risk under adversarial conditions
  If the upstream OIDC provider is slow or an attacker controls the discovery/JWKS endpoint (e.g., via a misconfigured `disco_doc_address`), the sync `threading.Lock` is held for the full HTTP timeout (default 30s x 3 retries = up to ~90s worst case). All other threads calling `_get_disco_response` or `_get_cached_jwks` for the same or different URIs block on the same global lock during that entire window. The previous pattern (fetch outside lock) was vulnerable to cache stampede but did not block all threads.
  Mitigation: Consider a per-URI lock (e.g., keyed lock dictionary) so that a slow fetch for one issuer does not block threads validating tokens from a different issuer. Alternatively, set a tighter per-request timeout for these fetches. This is classified UNLIKELY because it requires the upstream provider to be adversarial or severely degraded, and the library already has a 30s timeout.

### INFO (acceptable risk)

- `src/py_identity_model/aio/token_validation.py:65-78` — Async lock held during `await` is standard Python asyncio pattern. Unlike the sync case, other coroutines yield cooperatively while waiting on the lock; no thread starvation occurs. The async stampede fix is clean.
- `src/tests/unit/test_cache_stampede.py` — Test file directly accesses internal module state (`_jwks_cache`, `_disco_cache`) to simulate TTL expiry. This is test code only and does not affect production security posture. No secrets or credentials are present.

### Summary
- BLOCK: 0 | WARN: 1 | INFO: 2
- Overall: PASS

jamescrowley321:
## Review: Red Team (Viper)

### Attack Surface
- `src/py_identity_model/sync/token_validation.py` — Cache fetch pattern changed: discovery and JWKS HTTP fetches now execute while holding the cache lock (single-flight stampede prevention)
- `src/py_identity_model/aio/token_validation.py` — Same pattern change for async path: discovery and JWKS fetches moved inside `asyncio.Lock` scope
- `src/tests/unit/test_cache_stampede.py` — New test file (no production security impact)

### Findings

#### MEDIUM — Denial of Service via Lock Starvation (Sync: Blocking I/O Under threading.Lock)

- **Location**: `src/py_identity_model/sync/token_validation.py:65-78` (`_get_disco_response`), lines 96-107 (`_get_cached_jwks`), lines 112-119 (`_refresh_jwks`)
- **Attack scenario**:
  1. Attacker causes the upstream OIDC provider (or a DNS/network intermediary) to become slow or unresponsive (e.g., by flooding the provider, or via DNS poisoning to redirect to a tarpit).
  2. The first thread to hit an expired cache entry acquires `_disco_cache_lock` (or `_jwks_cache_lock`) and begins the HTTP fetch, which now executes entirely under the lock.
  3. The HTTP fetch includes retry logic with exponential backoff: up to 3 retries with delays of 1s, 2s, 4s, plus the 30-second timeout per attempt. Worst case: 4 attempts x 30s timeout + 7s backoff = ~127 seconds holding the lock.
  4. All other threads attempting to validate tokens block on the lock for the entire duration. In a web application (FastAPI sync workers, Flask, etc.), this blocks all request-handling threads, causing a complete service outage.
  5. Previously, the fetch happened outside the lock, so only the cache-write was serialized (microseconds). Now the entire network round-trip serializes all threads.
- **Prerequisites**: Attacker needs ability to degrade network connectivity to the OIDC provider (network position, DNS manipulation, or simply the provider experiencing an outage — no attacker action required for that scenario).
- **CVSS v3.1**: 5.9 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:H)
- **Remediation**: Implement a lock-with-timeout pattern. Acquire the lock with a bounded timeout (e.g., 5 seconds). If the lock cannot be acquired, either serve the stale cached response (if available) or raise a fast-fail error rather than blocking indefinitely. Alternatively, use a single-flight pattern that does not hold the lock during I/O: acquire the lock, set a "refresh in progress" sentinel, release the lock, perform the fetch, then re-acquire the lock to store the result while other waiters return the stale value or await a shared future/event.

#### LOW — Async Lock Held Across Await Points Prevents Concurrent Progress

- **Location**: `src/py_identity_model/aio/token_validation.py:65-78` (`_get_disco_response`), lines 96-107 (`_get_cached_jwks`), lines 112-119 (`_refresh_jwks`)
- **Attack scenario**:
  1. In the async path, `asyncio.Lock` is held while `await get_discovery_document(...)` and `await get_jwks(...)` execute, including retry/backoff `asyncio.sleep()` calls.
  2. If the upstream provider is slow or rate-limiting (429 responses trigger retries), the lock is held for the full retry cycle (up to ~127 seconds worst case).
  3. All other coroutines calling `_get_disco_response` or `_get_cached_jwks` for ANY URL are blocked, since there is a single global lock per cache type (not per-URL).
  4. Unlike the sync case, this does not block OS threads (asyncio is cooperative), so the event loop remains responsive for non-token-validation tasks. However, ALL token validation requests queue behind the slow fetch, causing token validation latency to spike to the full retry duration for every concurrent request.
  5. An attacker triggering rate-limiting (429) against the OIDC provider amplifies this: a single slow URL blocks validation for all URLs.
- **Prerequisites**: Ability to cause slow responses from the OIDC provider (network conditions, provider outage, or deliberate rate-limit triggering). The single global lock (not per-URI) is the amplifier.
- **Remediation**: Same as above — use a per-URI lock or a single-flight future pattern (e.g., `asyncio.Event` or a shared `asyncio.Task` per cache key) so that only one fetch per URI is in-flight, while other URIs proceed independently.

#### LOW — Cache Clear Functions Not Protected by Lock (Pre-existing, Amplified by Change)

- **Location**: `src/py_identity_model/sync/token_validation.py:81-83` (`clear_discovery_cache`), line 122-124 (`clear_jwks_cache`); same in `aio/token_validation.py:81-83`, lines 122-124
- **Attack scenario**:
  1. `clear_discovery_cache()` and `clear_jwks_cache()` call `.clear()` on the cache dict without acquiring the corresponding lock.
  2. If `clear_jwks_cache()` is called while another thread/coroutine is inside the lock performing a fetch-and-store, the `.clear()` races with the cache write. In CPython, `dict.clear()` is atomic due to the GIL, so this is not a memory-safety issue, but it can cause the freshly-fetched result to be immediately discarded, forcing a re-fetch on the next request.
  3. With the new pattern (fetch under lock), the race window is wider because the lock is held longer, making it more likely that a concurrent `clear()` call overlaps with an in-progress fetch.
  4. This is primarily a correctness issue rather than a direct security exploit, but in combination with the lock starvation issue above, an attacker who can trigger cache clears (e.g., via an admin API that exposes this) could force repeated slow fetches.
- **Prerequisites**: Ability to trigger `clear_discovery_cache()` or `clear_jwks_cache()` — typically requires application-level access (admin endpoint, test harness, or code injection). Not directly exploitable from the network.
- **CVSS v3.1**: 2.0 (AV:L/AC:H/PR:H/UI:N/S:U/C:N/I:N/A:L)
- **Remediation**: Acquire the cache lock before calling `.clear()` in the clear functions, or document that these functions are not safe to call concurrently with token validation.

### Summary
- Attack surface elements: 3 (sync token_validation cache lock, async token_validation cache lock, new test file)
- Overall: PASS (the stampede fix is a net security improvement; the medium finding is an availability trade-off that should be addressed but does not constitute a blocking security defect)

PR #391, opened 2026-05-10
fix(token-validation): refresh JWKS on cached kid miss + stampede pinning + hardening

jamescrowley321:
## Summary

Consolidates #390 (kid-miss rotation fix) into a single PR. Three commits:

- **72e6808** `fix(token-validation): refresh JWKS on cached kid miss` — original rotation fix from #390. On the cached JWKS path, when the incoming JWT's \`kid\` is not in the cached JWKS, force a refresh via \`_refresh_jwks\` instead of waiting for a \`SignatureVerificationException\` that never fires (\`find_key_by_kid\` raises before \`decode_with_config\` runs). Mirrored in sync + async \`_discover_and_resolve_key\`. DI (injected \`http_client\`) path unchanged. Required for the OIDF Config RP \`oidcc-client-test-signing-key-rotation\` conformance test.

- **6ceaa12** `test(token-validation): pin kid-miss stampede protection` — three tests in \`test_cache_stampede.py\` that prove the new pre-decode refresh path cannot fan out into N concurrent JWKS fetches under contention. 25 concurrent async tasks with the same unknown kid → exactly 2 fetches (1 prime + 1 refresh). Same with 10 distinct unknown kids (single-flight is keyed by \`jwks_uri\`, not kid). Same with 25 sync threads. Pins the existing protection (\`_jwks_cache_lock\` serialization + the \`cached_at >= request_time\` guard inside \`_refresh_jwks\`).

- **5a5a189** `fix(token-validation): harden kid-miss refresh path...` — three substantive improvements from the adversarial review:
  1. **Diagnostic warning** when refresh succeeds but the kid is still absent. Distinguishes rotation issues from configuration issues in production logs.
  2. **Fixed test priming** in \`test_cached_path_refreshes_jwks_when_kid_not_in_cache\` (sync + async). The prior \`contextlib.suppress\` warmup actually triggered the legacy retry path during priming, so the assertion passed without exercising the new code. Direct \`_get_cached_jwks\` priming makes the assertion meaningful.
  3. **Negative regression test** \`test_no_refresh_when_kid_present_in_cache\` (sync + async) — pins the \`not any(...)\` predicate sense; a future flip would be caught immediately.
  4. Replaced \`var, _ = generate_rsa_keypair()\` with \`var = generate_rsa_keypair()[0]\` across affected test files so unused PEMs aren't bound to local frame variables that pytest's \`--showlocals\` would dump on failure.

## What was considered and rejected

A \`KID_MISS_REFRESH_COOLDOWN\` to rate-limit attacker amplification, and a \`try/except\` around \`_refresh_jwks\` to fall through to stale cache on network failures. Both were overengineered:
- The amplification factor is 1:1, not 1:N — every request triggers one fetch with or without cooldown. Real DoS protection belongs at the edge / WAF / OP, not in a token-validation library masquerading as a rate limiter.
- The error fallback masked actual network errors with \"no matching kid found\", which is less diagnostic than the original behavior.
- The cooldown could only be tested by monkeypatching the constant — which proves nothing about production behavior at the production cooldown.

## Known follow-ups (not in this PR)

- HTTP cache header handling has real gaps: failed responses get cached for the default 24h, \`Cache-Control: no-store\` / \`no-cache\` are ignored, no \`Age\` subtraction, no \`ETag\`/conditional revalidation. Pre-existing — separate PR.

## Test plan

- [x] \`make lint\` — ruff + ruff format + pyrefly + pytest 80% coverage all pass
- [x] \`make test-unit\` — 993 tests pass locally
- [ ] Local conformance suite — confirm Config RP rotation test still PASSes

### DIFF
--- a/conformance/results/basic-rp-latest.json
+++ b/conformance/results/basic-rp-latest.json
@@ -1,104 +1,104 @@
 {
   "plan": "basic-rp",
-  "plan_id": "k14GyYkholz4l",
+  "plan_id": "T6YsG1bnx4c8E",
   "suite_url": "https://localhost.emobix.co.uk:8443",
   "results": [
     {
       "test": "oidcc-client-test",
       "status": "PASSED",
-      "test_id": "qAV7nbRtxkI4c40",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=qAV7nbRtxkI4c40",
+      "test_id": "J6RZ0QYLIqSqnmN",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=J6RZ0QYLIqSqnmN",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-invalid-iss",
       "status": "PASSED",
-      "test_id": "EtKLH4AXQe4Uh0n",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=EtKLH4AXQe4Uh0n",
+      "test_id": "9NiPTvBCVqm8Ual",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=9NiPTvBCVqm8Ual",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-missing-sub",
       "status": "PASSED",
-      "test_id": "dumfCr7PFyQP10b",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=dumfCr7PFyQP10b",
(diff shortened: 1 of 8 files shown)


jamescrowley321:
Superseded by #394, which combines this PR with #392 plus five additional adversarial-review fixes. Specifically: this PR's pre-decode kid-miss refresh shipped a 1:1 DoS amplification vector (no cooldown). #394 keeps the rotation fix and adds a 5-second per-jwks_uri cooldown bounding upstream amplification.

PR #392, opened 2026-05-11
fix(jwks-cache): respect HTTP cache headers (no-store, no-cache, failed-response handling)

jamescrowley321:
## Summary

Three correctness fixes to the JWKS and discovery cache, all confined to the cache layer. No behavior change for the happy path (successful response with plain `max-age`).

- **Failed responses are no longer cached.** Previously, when `get_jwks()` or `get_discovery_document()` returned `is_successful=False` (network error, 4xx after retries, etc.), the cache layer still stored the failed response under the default 24h TTL (JWKS) or 1h TTL (discovery). A transient blip could therefore poison the cache for up to a day. Now the cache layer checks `response.is_successful` and skips caching on failure, so the next call re-fetches.
- **`Cache-Control: no-store` is honored** (RFC 7234 §5.2.2.5). When present, the response is returned to the caller but not stored in the cache.
- **`Cache-Control: no-cache` is honored** (RFC 7234 §5.2.2.4). The simplest correct behavior is to skip caching and re-fetch on every call.

Implementation:
- Adds `is_uncacheable(cache_control)` in `core/jwks_cache.py` — case-insensitive regex match for `no-store` or `no-cache` directives.
- Gates the cache write in `_get_cached_jwks`, `_refresh_jwks`, and `_get_disco_response` (sync and async) on both `response.is_successful` and `not is_uncacheable(response.cache_control)`.

Deferred (lower priority, not implemented):
- `Age` header subtraction from effective TTL (matters mainly behind CDNs).
- `Expires` header (HTTP/1.0 fallback) parsing — modern OPs send `Cache-Control`.
- `ETag` / `If-None-Match` conditional revalidation — bandwidth optimization, not a correctness bug.

## Test plan

- [x] New unit tests in `test_cache_http_headers.py` cover all three fixes for both sync and async paths:
  - 404 → cache miss; subsequent fetch succeeds and populates the cache.
  - `no-store, max-age=600` → response returned, cache empty.
  - `no-cache, max-age=600` → every call hits the origin (3-call assertion).
- [x] New unit tests for `is_uncacheable()` covering no-store, no-cache, combined directives, case-insensitivity, and the non-matching cases (plain `max-age`, `None`, `""`).
- [x] `make lint` passes (ruff lint, ruff format, pyrefly, coverage gate at 95.55%).
- [x] `make test-unit` passes (1006 tests).

### DIFF
--- a/src/py_identity_model/aio/token_validation.py
+++ b/src/py_identity_model/aio/token_validation.py
@@ -13,6 +13,7 @@ from ..core.jwks_cache import (
     DiscoCacheEntry,
     JwksCacheEntry,
     is_cache_expired,
+    is_uncacheable,
     resolve_disco_ttl,
     resolve_ttl,
 )
@@ -76,10 +77,11 @@ async def _get_disco_response(
         response = await get_discovery_document(
             DiscoveryDocumentRequest(address=disco_doc_address, policy=policy),
         )
-        ttl = resolve_disco_ttl(response.cache_control)
-        _disco_cache[cache_key] = DiscoCacheEntry(
-            response=response, cached_at=time.time(), ttl=ttl
-        )
+        if response.is_successful and not is_uncacheable(response.cache_control):
+            ttl = resolve_disco_ttl(response.cache_control)
+            _disco_cache[cache_key] = DiscoCacheEntry(
+                response=response, cached_at=time.time(), ttl=ttl
+            )
         return response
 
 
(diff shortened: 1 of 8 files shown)


jamescrowley321:
Superseded by #394, which combines this PR with #391 plus five additional adversarial-review fixes. Critically: this PR's _refresh_jwks change introduced a regression — when the refreshed response carried Cache-Control: no-store/no-cache (which Descope sends), the stale entry was left in the cache, causing every subsequent token validation to re-trigger _refresh_jwks until the original 24h TTL expired. #394 fixes that by routing through apply_jwks_cache_outcome which pops the stale entry on uncacheable refresh.

PR #394, opened 2026-05-16
fix(jwks-cache): combined PR #391+#392 with adversarial-review hardening

jamescrowley321:
## Summary

Combines and supersedes #391 (kid-miss refresh + stampede pinning) and #392 (HTTP cache headers). Adds five additional fixes from an adversarial review of the cache primitive, every fix with a proving test that fails against the pre-fix code.

**Headline:** PR #392 as drafted introduced a regression — when a refresh response carried `Cache-Control: no-store` or `no-cache` (which Descope sends on JWKS), the fresh response was returned to the caller but the **stale entry was left in the cache**. Every subsequent token validation hit the stale entry, re-triggered the refresh, and hammered the upstream JWKS endpoint until the original 24h TTL expired — the exact DoS the cache exists to prevent. The first commit on this branch (after the two original PR commit chains) fixes that.

## Inherited from #391

- `fix(token-validation): refresh JWKS on cached kid miss` — pre-decode refresh on kid-miss in the cached path
- `test(token-validation): pin kid-miss stampede protection`
- `fix(token-validation): harden kid-miss refresh path with diagnostic, fixed test priming, and negative regression`

## Inherited from #392

- `fix(jwks-cache): respect HTTP cache headers for no-store, no-cache, and failed responses`
- `test(integration): skip caching benchmarks when provider sends no-store/no-cache`

### C1 + H3 — `fix(jwks-cache): invalidate stale entry on uncacheable refresh and reject empty keys`

Decision-matrix helper (`apply_jwks_cache_outcome` / `apply_disco_cache_outcome`) in `core/jwks_cache.py`:

- Unsuccessful: retain existing (jose4j's retain-on-error semantics).
- `no-store`/`no-cache`: **pop** existing — fixes the regression that PR #392 introduced.
- Empty `keys: []`: retain existing — empty JWKS is a transient blip, never a valid replacement.
- Successful + cacheable + non-empty: store.

Sync and async share the helper so the failure-mode matrix cannot drift.

### H4 + C3 + C4 — `fix(jwks-cache): cap DoS amplification, clamp env config, bound cache size`

- **H4 kid-miss cooldown** (default 5s, env-tunable `KID_MISS_REFRESH_COOLDOWN`). Bounds the 1:1 amplification PR #391 introduced. Design borrows from Microsoft IdentityModel's `RefreshInterval` and Auth0 `node-jwks-rsa`'s `rateLimit`. Per-jwks_uri so an attacker against OP-A can't suppress refreshes for OP-B. Bypassed when cache is empty (cold start).
- **C3 env TTL clamping**. `JWKS_CACHE_TTL=0` (instantly-expired cache → DoS) and `JWKS_CACHE_TTL=2592000` (silent override of documented 24h ceiling) are now both clamped to `[60s, 86400s]`. Garbage values fall back to default with a warning instead of crashing.
- **C4 bounded LRU**. `JWKS_CACHE_MAX_ENTRIES` (default 64, env-tunable) caps memory under multi-tenant gateway scenarios. FIFO eviction via Python's insertion-ordered dicts; `apply_*_cache_outcome` pop-and-reinsert so refreshed URIs move to newest.

### C2 — `fix(jwks-cache): use per-URI lock striping so distinct issuers fetch in parallel`

Pre-fix: single `threading.Lock`/`asyncio.Lock` held across the upstream HTTP fetch. A fetch to `https://idp-a/jwks` blocked every other fetch to any other address. Under async this was strictly worse — held across `await`, the lock owned the event loop.

Post-fix: cache-aside with 32-stripe per-URI lock hash. Cheap `dict.get()` returns fresh entries with no lock at all. Only the upstream fetch on miss serializes — per URI. Same-URI single-flight semantics preserved.

## Proving tests

Every fix has a test that **fails against pre-fix code** and **passes against the fix**. I verified this for C1/H3 and H4 by stashing the fix and running the test against the pre-fix code:

| Fix | Test file | Proves |
|---|---|---|
| C1 + H3 | `test_cache_http_headers.py` | 17 new tests across 4 classes; refresh-with-uncacheable-response and empty-keys-200 cases. |
| H4 | `test_kid_miss_cooldown.py` | 13 tests: 25-attacker-request amplification, cold-start bypass, rotation-after-cooldown propagation, per-jwks_uri isolation, env override. |
| C3 + C4 | `test_cache_env_and_bounds.py` | 27 tests: 8 parametrized TTL clamp cases × 2 (JWKS + disco), garbage-value fail-safe, max-entries env override, FIFO order preservation across refresh+overflow, end-to-end size cap. |
| C2 | `test_per_uri_locks.py` | 4 tests: distinct-URI parallel (sync + async, ~2x speedup proved by wall-clock), same-URI single-flight preserved. |

## What's deferred (separate PR / follow-up)

- **M2 monotonic clock** — `time.time()` → `time.monotonic()` for cache age math. Real bug under NTP slew but lower impact and bigger blast radius (touches several existing tests).
- **H5 documentation** — README + docstring note that single-flight is per-process; under prefork (gunicorn workers) each worker has its own cache.
- **H6 `http_client` cache bypass** — when callers inject their own client, the cache is bypassed entirely. Documented in docstring; a proper fix means restructuring the cache as a transport wrapper.

These three are tracked but not blockers for this PR.

## How this is safer than #391 + #392 alone

#391 alone shipped a 1:1 DoS amplifier (no cooldown). #392 alone shipped the no-store stale-cache regression (the headline finding above). Both gaps are closed here, plus four structural bugs in the underlying cache primitive (env clamping, bounded growth, per-URI locks, decision-matrix unification).

## Test plan

- [x] `make lint` passes (ruff + ruff format + pyrefly + 80% coverage gate at 95.75%).
- [x] `make test-unit` — 1076 tests pass.
- [ ] Local conformance suite — confirm OIDF Config RP `oidcc-client-test-signing-key-rotation` still PASSes (covered by CI).
- [ ] Integration tests against Descope (CI).

Closes #391
Closes #392

### DIFF
--- a/conformance/results/basic-rp-latest.json
+++ b/conformance/results/basic-rp-latest.json
@@ -1,104 +1,104 @@
 {
   "plan": "basic-rp",
-  "plan_id": "k14GyYkholz4l",
+  "plan_id": "T6YsG1bnx4c8E",
   "suite_url": "https://localhost.emobix.co.uk:8443",
   "results": [
     {
       "test": "oidcc-client-test",
       "status": "PASSED",
-      "test_id": "qAV7nbRtxkI4c40",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=qAV7nbRtxkI4c40",
+      "test_id": "J6RZ0QYLIqSqnmN",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=J6RZ0QYLIqSqnmN",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-invalid-iss",
       "status": "PASSED",
-      "test_id": "EtKLH4AXQe4Uh0n",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=EtKLH4AXQe4Uh0n",
+      "test_id": "9NiPTvBCVqm8Ual",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=9NiPTvBCVqm8Ual",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-missing-sub",
       "status": "PASSED",
-      "test_id": "dumfCr7PFyQP10b",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=dumfCr7PFyQP10b",
+      "test_id": "3HXnY2MWHR5ZpOr",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=3HXnY2MWHR5ZpOr",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-invalid-aud",
       "status": "PASSED",
-      "test_id": "c5kOspGFCSX3L2v",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=c5kOspGFCSX3L2v",
+      "test_id": "VITlE4esdcuZwts",
+      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=VITlE4esdcuZwts",
       "detail": ""
     },
     {
       "test": "oidcc-client-test-missing-iat",
       "status": "PASSED",
-      "test_id": "5BwxiEdKzRHEbbx",
-      "log_url": "https://localhost.emobix.co.uk:8443/log-detail.html?log=5BwxiEdKzRHEbbx",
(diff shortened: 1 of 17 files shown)


jamescrowley321:
Superseded by #395, which contains these three commits (`86b5d12` C1/H3, `d9dfde2` H4/C3/C4, `c6cf4d5` C2) plus the second-review hardening (`616f5d7`) and the Makefile conformance race fix (`b901cea`). Closing in favor of the consolidated PR.

Issue #213, opened 2026-03-24
Implement JWT Client Authentication (private_key_jwt / client_secret_jwt)


jamescrowley321:
## Goal

Implement RFC 7523 JWT Bearer Client Authentication for OAuth 2.0 token endpoint, supporting both `private_key_jwt` and `client_secret_jwt` methods.

## Context

Currently, py-identity-model only supports `client_secret_post` and `client_secret_basic` authentication. JWT-based client authentication is **required** for FAPI 2.0 compliance (#97) and is a best practice for confidential clients that want to avoid transmitting secrets.

## Specification References

- **RFC 7523** — [JWT Profile for OAuth 2.0 Client Authentication](https://datatracker.ietf.org/doc/html/rfc7523)
  - §2.2 Using JWTs for Client Authentication
  - Client assertion JWT claims: `iss` (client_id), `sub` (client_id), `aud` (token endpoint), `jti`, `exp`, `iat`
  - `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer`
  - `client_assertion` — The signed JWT
- **OpenID Connect Core 1.0** — §9 Client Authentication
  - `private_key_jwt` — Client signs JWT with its private key; AS verifies with registered public key
  - `client_secret_jwt` — Client signs JWT with client_secret via HMAC
- **OAuth 2.1** (draft-ietf-oauth-v2-1) — §2.4 Client Authentication — References RFC 7523
- **RFC 8414** — §2 — `token_endpoint_auth_methods_supported` metadata
- **FAPI 2.0** — Requires `private_key_jwt` or mTLS; `client_secret_basic`/`client_secret_post` are NOT sufficient

## Duende IdentityModel Reference

- [`ClientAssertion`](https://github.com/DuendeSoftware/foss/blob/main/identity-model/src/IdentityModel/Client/Messages/ClientAssertion.cs) — Model with `Type` and `Value`
- [`ProtocolRequest`](https://github.com/DuendeSoftware/foss/blob/main/identity-model/src/IdentityModel/Client/Messages/ProtocolRequest.cs) — `ClientAssertion` property, `ClientCredentialStyle` enum
- [`JwtClientAssertion`](https://github.com/DuendeSoftware/foss/blob/main/identity-model/src/IdentityModel/Client/Messages/ClientAssertion.cs) — JWT assertion creation helper

## Tasks

- [ ] Create `ClientAssertion` model (type + value)
[...]

Issue #88, opened 2025-10-21
Create Base Request/Response Classes


jamescrowley321:
**Suggested Labels**: refactoring, enhancement

## Goal
Establish base classes for consistent request/response handling across all endpoints.

## Tasks
- [ ] Create `BaseRequest` abstract class
- [ ] Create `BaseResponse` abstract class with `is_successful` and `error` pattern
- [ ] Refactor existing requests to inherit from `BaseRequest`
  - DiscoveryDocumentRequest
  - JwksRequest
  - ClientCredentialsTokenRequest
- [ ] Refactor existing responses to inherit from `BaseResponse`
  - DiscoveryDocumentResponse
  - JwksResponse
  - ClientCredentialsTokenResponse
- [ ] Add common error handling patterns
- [ ] Add validation helpers
- [ ] Update documentation

## Success Criteria
- Consistent API across all request/response types
- Reduced code duplication
- Better error handling
- Easier to add new endpoints

**Milestone**: v0.1.0 - Foundation

PR #433, opened 2026-06-30
feat(client-auth): add private_key_jwt (RFC 7523) client authentication

jamescrowley321:
## Summary
- Add **private_key_jwt** (RFC 7523 / OIDC Core §9) client authentication across every client-authenticating endpoint: token (auth-code, refresh, client-credentials), token-exchange, device-auth, device-token, PAR, introspection, revocation.
- New `core/client_assertion.py` — `build_client_assertion()` + `apply_private_key_jwt()`; injects `client_assertion` + `client_assertion_type` into the request body and returns `auth=None` (no Basic header).
- New `core/jwt_signing.py` — shared `SUPPORTED_SIGNING_ALGORITHMS` + `validate_signing_algorithm()`, extracted from `core/jar.py` (now reuses it; no duplication). FAPI 2.0 algorithms: **PS256/ES256**.
- New `PrivateKeyJwt` dataclass (`private_key`, `algorithm='PS256'`, `kid`, `audience`, `lifetime=300`); optional `private_key_jwt` field added to all 9 request models. Exported from sync + aio + root `__init__`.
- Precedence: **private_key_jwt > client_secret (Basic) > public (client_id in body)**.
- Normalized `prepare_token_request_data` (client-credentials) to the 3-tuple `(params, headers, auth)` contract; sync/aio wrappers consume `auth` instead of hardcoding Basic.
- Assertion claims `iss=sub=client_id`, `aud` (defaults to request address, overridable), `jti` (uuid4), `iat`/`nbf` backdated 10s for clock skew, `exp = iat + lifetime`.

Refs #213

## Review Findings Addressed
5 independent reviewers (Blind / Edge / Acceptance / Sentinel / Viper).
- **P0 (Blind + Edge, 14 sites):** `prepare_*_request_data()` now signs the assertion (pyjwt.encode), so a malformed PEM/key-alg mismatch raised uncaught outside the wrappers' try/except. **Fixed** — moved every sync+aio `prepare_*` call inside the existing try/except so signing failures route through the structured error handlers (regression tests added, sync + async).
- **P1 clock skew (Blind):** **Fixed** — backdate `iat`/`nbf` by a 10s leeway; `exp = iat + lifetime` preserved.
- **P1 fixture coupling (Blind):** **Fixed** — added a unit-tier self-check deriving the checked-in key's public coords and asserting equality with the provider.js JWK (fails loud on drift).
- **Security: Sentinel PASS, Viper PASS** (0 crit/high/med; 3 LOW informational, no change needed).
- **Acceptance 4 FAILs** (`client_secret_jwt`, `ClientAssertion(type,value)` model, FastAPI example) are **out of scope** per the issue's FAPI2 = private_key_jwt PS256/ES256 design — scope disagreements, not defects.

## Test plan
- [x] Unit tests pass (`src/tests/unit/test_client_assertion.py` — sync + async via respx, all-endpoint precedence, claim/signature/alg-validation/jti, bad-key→error, iat/nbf backdating, key↔JWK self-check)
- [x] Integration tests pass (`src/tests/integration/test_private_key_jwt.py` against node-oidc-provider: client-credentials, PAR, full auth-code exchange, adversarial wrong-key rejection — full suite **115 passed / 5 skipped**)
- [x] Lint passes (ruff, ruff format, pyrefly, 80% coverage)
- [ ] CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ uv add py-identity-model
 | DPoP | RFC 9449 | Done |
 | Pushed Authorization Requests | RFC 9126 | Done |
 | JWT Secured Authorization Request | RFC 9101 | Done |
+| private_key_jwt Client Authentication | RFC 7523 | Done |
 | FAPI 2.0 Security Profile | FAPI 2.0 | Done |
 | UserInfo Endpoint | OIDC Core | Done |
 | OAuth Callback State Validation | RFC 6749 | Done |
@@ -104,6 +105,7 @@ Each protocol feature has a standalone example in [`examples/`](examples/):
 | [dpop_example.py](examples/dpop_example.py) | DPoP proof creation (RFC 9449) |
 | [par_example.py](examples/par_example.py) | Pushed Authorization Requests (RFC 9126) |
 | [jar_example.py](examples/jar_example.py) | JWT Secured Authorization Request (RFC 9101) |
+| [private_key_jwt_example.py](examples/private_key_jwt_example.py) | private_key_jwt client authentication (RFC 7523) |
 | [fapi_example.py](examples/fapi_example.py) | FAPI 2.0 Security Profile |
 | [device_auth_example.py](examples/device_auth_example.py) | Device Authorization Grant (RFC 8628) |
 | [token_exchange_example.py](examples/token_exchange_example.py) | Token Exchange (RFC 8693) |
--- a/examples/private_key_jwt_example.py
+++ b/examples/private_key_jwt_example.py
@@ -1,0 +1,128 @@
+"""
+private_key_jwt Client Authentication Example (RFC 7523)
+
+Demonstrates authenticating a confidential client to the token endpoint with
+a signed JWT assertion (``private_key_jwt``) instead of a client secret, as
+required by FAPI 2.0 and recommended by OAuth 2.1 / OpenID Connect Core 1.0
+Section 9.
+
+Flow:
+1. Generate an EC P-256 (ES256) key pair and serialize the private key to PEM.
+2. Discover ``token_endpoint_auth_methods_supported`` to confirm the
+   authorization server accepts ``private_key_jwt``.
+3. Build a ``PrivateKeyJwt`` and attach it to a token request.
+4. Observe that the assertion (not Basic auth, not client_secret) is used.
+"""
+
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec
+
+from py_identity_model import (
+    ClientCredentialsTokenRequest,
+    DiscoveryDocumentRequest,
+    PrivateKeyJwt,
+    get_discovery_document,
+)
+
+
+# A placeholder issuer; swap for your own provider when running for real.
+DEMO_DISCOVERY_URL = "https://demo.duendesoftware.com/.well-known/openid-configuration"
+
+
+def generate_es256_private_key_pem() -> str:
+    """Generate an EC P-256 key pair and return the private key as PEM.
+
+    ES256 (ECDSA on the NIST P-256 curve) is one of the two signing
+    algorithms permitted by FAPI 2.0 for ``private_key_jwt`` (the other is
+    ``PS256``). In production the matching public key (JWK) is registered with
+    the authorization server so it can verify the assertion signature.
+    """
+    private_key = ec.generate_private_key(ec.SECP256R1())
+    pem = private_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+    return pem.decode("utf-8")
+
+
+def discover_auth_methods_example():
+    """Discover the auth methods supported by the authorization server."""
+    print("\n" + "=" * 60)
+    print("Discover token_endpoint_auth_methods_supported")
+    print("=" * 60)
+
+    response = get_discovery_document(
+        DiscoveryDocumentRequest(address=DEMO_DISCOVERY_URL)
+    )
+
+    if not response.is_successful:
+        print(f"  Discovery failed: {response.error}")
+        return None
+
+    methods = response.token_endpoint_auth_methods_supported or []
+    print(f"\n  Token endpoint: {response.token_endpoint}")
+    print(f"  Supported auth methods: {methods}")
+
+    if "private_key_jwt" in methods:
+        print("  -> Server supports private_key_jwt")
+    else:
+        print("  -> Server did NOT advertise private_key_jwt")
+
+    return response.token_endpoint
+
+
+def private_key_jwt_token_request(token_endpoint: str):
+    """Build a client credentials request authenticated with private_key_jwt."""
+    print("\n" + "=" * 60)
+    print("Client Credentials with private_key_jwt")
+    print("=" * 60)
+
+    # Step 1: Generate (or load) the signing key.
+    private_key_pem = generate_es256_private_key_pem()
+    print("\n  Generated EC P-256 (ES256) private key (PEM).")
+
+    # Step 2: Build the private_key_jwt authentication parameters.
+    #   - algorithm "ES256" matches the key type (FAPI 2.0 compliant).
+    #   - audience defaults to the request address (token endpoint) when None.
+    #   - kid is optional; set it when the AS needs help locating your key.
+    auth = PrivateKeyJwt(private_key=private_key_pem, algorithm="ES256")
+
+    # Step 3: Attach the assertion to a token request. Note that a
+    # client_secret is also provided here only to demonstrate precedence:
+    # when private_key_jwt is set, it WINS and the client_secret is ignored,
+    # so no HTTP Basic (client_secret_basic) header is sent.
+    token_request = ClientCredentialsTokenRequest(
+        address=token_endpoint,
+        client_id="my-confidential-client",
+        client_secret="ignored-because-private-key-jwt-takes-precedence",
+        scope="api",
+        private_key_jwt=auth,
+    )
+
+    print(f"  Token endpoint: {token_request.address}")
+    print(f"  Client ID: {token_request.client_id}")
+    print(f"  Assertion algorithm: {auth.algorithm}")
+    print(f"  private_key_jwt set: {token_request.private_key_jwt is not None}")
+    print("  client_secret present but IGNORED (private_key_jwt wins).")
+    print("  No Authorization: Basic header is sent; the request body carries")
+    print("    client_assertion_type + client_assertion instead.")
+    print("  (Would call request_client_credentials_token(token_request) here)")
+
+
+def main():
+    print("\n" + "=" * 60)
+    print("private_key_jwt CLIENT AUTHENTICATION (RFC 7523)")
+    print("=" * 60)
+
+    token_endpoint = discover_auth_methods_example()
+    # Fall back to a placeholder endpoint when discovery is unavailable.
+    private_key_jwt_token_request(token_endpoint or "https://auth.example.com/token")
+
+    print("\n" + "=" * 60)
+    print("Examples completed!")
+    print("=" * 60)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/py_identity_model/__init__.py
+++ b/src/py_identity_model/__init__.py
@@ -57,6 +57,7 @@ from .sync import (
     JsonWebKeyParameterNames,
     JwksRequest,
     JwksResponse,
+    PrivateKeyJwt,
     PushedAuthorizationRequest,
     PushedAuthorizationResponse,
     RefreshTokenRequest,
@@ -155,6 +156,8 @@ __all__ = [
     "JwksRequest",
     "JwksResponse",
     "NetworkException",
+    # Client authentication (private_key_jwt)
+    "PrivateKeyJwt",
     # PAR
     "PushedAuthorizationRequest",
     "PushedAuthorizationResponse",
--- a/src/py_identity_model/aio/__init__.py
+++ b/src/py_identity_model/aio/__init__.py
@@ -49,7 +49,7 @@ from ..core.fapi import (
     validate_fapi_discovery,
 )
 from ..core.jar import build_jar_authorization_url, create_request_object
-from ..core.models import BaseRequest, BaseResponse
+from ..core.models import BaseRequest, BaseResponse, PrivateKeyJwt
 from ..core.pkce import (
     generate_code_challenge,
     generate_code_verifier,
@@ -159,6 +159,7 @@ __all__ = [
     "JwksRequest",
     "JwksResponse",
     # PAR
+    "PrivateKeyJwt",
     "PushedAuthorizationRequest",
     "PushedAuthorizationResponse",
     # Refresh Token
--- a/src/py_identity_model/aio/device_auth.py
+++ b/src/py_identity_model/aio/device_auth.py
@@ -46,10 +46,10 @@ async def request_device_authorization(
         and ``verification_uri`` for user display.
     """
     log_device_auth_request(request)
-    params, headers, auth = prepare_device_auth_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_device_auth_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _request_device_auth(
             client, request.address, params, headers, auth
@@ -89,10 +89,10 @@ async def poll_device_token(
         indicating poll status.
     """
     log_device_token_request(request)
-    params, headers, auth = prepare_device_token_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_device_token_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _poll_device_token(
             client, request.address, params, headers, auth
--- a/src/py_identity_model/aio/introspection.py
+++ b/src/py_identity_model/aio/introspection.py
@@ -47,10 +47,10 @@ async def introspect_token(
         TokenIntrospectionResponse with claims dict including ``active`` bool.
     """
     log_introspection_request(request)
-    params, headers, auth = prepare_introspection_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_introspection_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _introspect_token(
             client, request.address, params, headers, auth
--- a/src/py_identity_model/aio/revocation.py
+++ b/src/py_identity_model/aio/revocation.py
@@ -47,10 +47,10 @@ async def revoke_token(
         TokenRevocationResponse indicating success or error.
     """
     log_revocation_request(request)
-    params, headers, auth = prepare_revocation_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_revocation_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _revoke_token(client, request.address, params, headers, auth)
         return process_revocation_response(response)
--- a/src/py_identity_model/aio/token_client.py
+++ b/src/py_identity_model/aio/token_client.py
@@ -70,10 +70,10 @@ async def request_client_credentials_token(
         ClientCredentialsTokenResponse: Token response
     """
     log_token_request(request)
-    params, headers, auth = prepare_token_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_token_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _request_token(client, request.address, params, headers, auth)
         return process_token_response(response)
@@ -98,10 +98,10 @@ async def request_authorization_code_token(
         AuthorizationCodeTokenResponse with token dict or error.
     """
     log_auth_code_token_request(request)
-    params, headers, auth = prepare_auth_code_token_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_auth_code_token_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _request_token(client, request.address, params, headers, auth)
         return process_auth_code_token_response(response)
@@ -126,10 +126,10 @@ async def refresh_token(
         RefreshTokenResponse with new token dict or error.
     """
     log_refresh_token_request(request)
-    params, headers, auth = prepare_refresh_token_request_data(request)
 
     response = None
     try:
+        params, headers, auth = prepare_refresh_token_request_data(request)
         client = http_client.client if http_client else get_async_http_client()
         response = await _request_token(client, request.address, params, headers, auth)
         return process_refresh_token_response(response)
--- a/src/py_identity_model/aio/token_client.py
+++ b/src/py_identity_model/aio/token_client.py
@@ -70,18 +70,12 @@ async def request_client_credentials_token(
         ClientCredentialsTokenResponse: Token response
     """
     log_token_request(request)
-    params, headers = prepare_token_request_data(request)
+    params, headers, auth = prepare_token_request_data(request)
 
     response = None
     try:
         client = http_client.client if http_client else get_async_http_client()
-        response = await _request_token(
-            client,
-            request.address,
-            params,
-            headers,
-            (request.client_id, request.client_secret),
-        )
+        response = await _request_token(client, request.address, params, headers, auth)
         return process_token_response(response)
     except Exception as e:
         return handle_token_error(e)
--- a/src/py_identity_model/core/client_assertion.py
+++ b/src/py_identity_model/core/client_assertion.py
@@ -70,15 +70,23 @@ def build_client_assertion(  # noqa: PLR0913  # RFC 7523 assertion params are al
         ("audience", audience),
         ("private_key", private_key),
     ):
-        if not value:
+        # Reject whitespace-only string values: they are truthy but would be
+        # signed into ``iss``/``sub``/``aud`` and rejected opaquely by the AS.
+        if not value or (isinstance(value, str) and not value.strip()):
             raise ValueError(f"{name} must not be empty")
 
     validate_signing_algorithm(algorithm, context="client assertion")
 
-    if lifetime <= 0:
-        raise ValueError(f"lifetime must be positive, got {lifetime}")
+    # ``exp`` is derived from the backdated ``iat`` (``now - leeway``), so a
+    # lifetime at or below the leeway yields an assertion already expired at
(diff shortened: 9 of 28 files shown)


jamescrowley321:
## Review: review-blind

## Review: Blind Hunter

### MUST FIX
- [`src/py_identity_model/sync/token_client.py:73`] `prepare_token_request_data(request)` is called OUTSIDE the `try/except` block. It now invokes `apply_private_key_jwt` -> `build_client_assertion` -> `pyjwt.encode`, which raises on an invalid/malformed PEM key or a key/algorithm mismatch. The exception propagates uncaught to the caller instead of being converted to an error response via `handle_token_error`, breaking the library's documented "return a response object" contract for the client_credentials flow.
- [`src/py_identity_model/sync/token_client.py:101`] `prepare_auth_code_token_request_data(request)` is called outside the `try/except`. A `private_key_jwt` signing failure (`pyjwt.encode`) raises an unhandled exception instead of being routed through `handle_auth_code_token_error`. Same crash-instead-of-error-response regression on the authorization_code flow.
- [`src/py_identity_model/sync/token_client.py:129`] `prepare_refresh_token_request_data(request)` is called outside the `try/except`; a `private_key_jwt` signing exception escapes uncaught instead of going through `handle_refresh_token_error`. Same regression on the refresh_token flow.
- [`src/py_identity_model/sync/introspection.py:50`] `prepare_introspection_request_data(request)` runs before the `try` block; a `private_key_jwt` `pyjwt.encode` failure crashes the call rather than returning an introspection error response.
- [`src/py_identity_model/sync/revocation.py:50`] `prepare_revocation_request_data(request)` runs before the `try` block; a signing failure escapes uncaught instead of producing a revocation error response.
- [`src/py_identity_model/sync/device_auth.py:49`] / [`src/py_identity_model/sync/device_auth.py:94`] `prepare_device_auth_request_data` / `prepare_device_token_request_data` run before the `try`; a `private_key_jwt` signing failure crashes instead of returning an error response.
- [`src/py_identity_model/aio/token_client.py:101`] / [`src/py_identity_model/aio/token_client.py:129`], [`src/py_identity_model/aio/introspection.py:50`], [`src/py_identity_model/aio/revocation.py:50`], [`src/py_identity_model/aio/device_auth.py:49`] / [`src/py_identity_model/aio/device_auth.py:92`] Same defect mirrored in the async API: the `prepare_*` call that now performs JWT signing sits outside `try/except`, so a signing exception propagates uncaught. (Note: PAR and token_exchange in both sync/async DO call prepare inside the `try` — the handling is inconsistent across endpoints, which is itself the smell.)

### SHOULD FIX
- [`src/py_identity_model/core/client_assertion.py:152`] The assertion `iat`/`exp` are derived from `int(time.time())` with no clock-skew margin. With `lifetime=300`, an authorization server whose clock is slightly ahead may reject a freshly minted assertion as not-yet-valid or already-expired at the boundary. Consider backdating `iat`/`nbf` by a few seconds. (Observed in the diff; standard hardening for issued JWTs.)
- [`src/py_identity_model/core/models.py:454`] Loosening `ClientCredentialsTokenRequest.client_secret`/`scope` to `None` defaults plus the sync/async change from always sending Basic `(client_id, client_secret)` to the `auth` returned by `prepare_token_request_data` silently changes behavior: a request with no secret and no `private_key_jwt` now sends `client_id` in the body as a public client instead of sending `Authorization: Basic client_id:None`. This is a public-vs-confidential auth-mode change for existing callers that omit the secret; verify no caller relied on the old (broken) Basic-with-None behavior.
- [`src/tests/integration/test_private_key_jwt.py:826`] / [`test-fixtures/node-oidc-provider/provider.js:1500`] Correct operation hinges on the checked-in ES256 private key's public coordinates exactly matching the hardcoded JWK `x`/`y` in `provider.js`. Nothing in the diff derives/asserts this match; a copy-paste mismatch would make every live test silently `skip`-or-fail with an opaque signature error. A unit assertion that the public key derived from `PKJWT_PRIVATE_KEY` equals the registered JWK would make the coupling self-checking.

### NITPICK
- [`src/py_identity_model/core/client_assertion.py:149`] `if kid is not None and not kid:` rejects empty-string `kid` but the public `PrivateKeyJwt.kid` field has no equivalent guard at construction; the validation lives only in the builder, so an empty `kid` is only caught deep in request preparation.
- [`src/py_identity_model/core/client_assertion.py:141`] `if not value:` for `private_key` treats an empty `bytes` (`b""`) and empty `str` identically as "empty", which is fine, but a whitespace-only PEM string passes this check and only fails later inside `pyjwt.encode` — see the MUST FIX items about where that failure surfaces.

jamescrowley321:
## Review: review-edge-case

## Review: Edge Case Hunter

### Findings

| `src/py_identity_model/sync/token_client.py:73` | Malformed PEM or key/algorithm mismatch in `private_key_jwt` | move `prepare_token_request_data(request)` inside the `try:` block | [CRASH] `pyjwt.encode` raises `ValueError`/`InvalidKeyError` uncaught, propagates to caller |
| `src/py_identity_model/aio/token_client.py:73` | Malformed PEM or key/algorithm mismatch in `private_key_jwt` | move `prepare_token_request_data(request)` inside the `try:` block | [CRASH] assertion build raises before try; exception escapes `request_client_credentials_token` |
| `src/py_identity_model/sync/token_client.py:101` | Bad signing key in auth-code `private_key_jwt` | move `prepare_auth_code_token_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught (prepare is outside try) |
| `src/py_identity_model/aio/token_client.py:101` | Bad signing key in auth-code `private_key_jwt` | move `prepare_auth_code_token_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |
| `src/py_identity_model/sync/token_client.py:129` | Bad signing key in refresh `private_key_jwt` | move `prepare_refresh_token_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught |
| `src/py_identity_model/aio/token_client.py:129` | Bad signing key in refresh `private_key_jwt` | move `prepare_refresh_token_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |
| `src/py_identity_model/sync/introspection.py:50` | Bad signing key in introspection `private_key_jwt` | move `prepare_introspection_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught |
| `src/py_identity_model/aio/introspection.py:50` | Bad signing key in introspection `private_key_jwt` | move `prepare_introspection_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |
| `src/py_identity_model/sync/revocation.py:50` | Bad signing key in revocation `private_key_jwt` | move `prepare_revocation_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught |
| `src/py_identity_model/aio/revocation.py:50` | Bad signing key in revocation `private_key_jwt` | move `prepare_revocation_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |
| `src/py_identity_model/sync/device_auth.py:49` | Bad signing key in device-auth `private_key_jwt` | move `prepare_device_auth_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught |
| `src/py_identity_model/sync/device_auth.py:94` | Bad signing key in device-token `private_key_jwt` | move `prepare_device_token_request_data(request)` inside `try:` | [CRASH] `pyjwt.encode` error escapes uncaught |
| `src/py_identity_model/aio/device_auth.py:49` | Bad signing key in device-auth `private_key_jwt` | move `prepare_device_auth_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |
| `src/py_identity_model/aio/device_auth.py:92` | Bad signing key in device-token `private_key_jwt` | move `prepare_device_token_request_data(request)` inside `try:` | [CRASH] assertion build raises before try; exception escapes |

### Summary
- Unhandled paths found: 14
- Critical (crash/data loss): 14

jamescrowley321:
## Review: review-acceptance

## Review: Acceptance Auditor

Scope note: this review covers the changes in `review-diff.patch` (commits `c136f68` feat + `9c04aa2` test for `private_key_jwt`). The branch contains additional unrelated files (`http_utils.py`, `test_retry_after.py`, http_client retry changes) that are NOT in the review diff and are out of scope for this AC audit.

The spec (issue #213) is a two-method feature: `private_key_jwt` AND `client_secret_jwt`. The diff implements only `private_key_jwt`. ACs are derived from the spec's Tasks, Integration Test Requirements, Example Requirements, and Success Criteria.

### PASS
- [AC-2] JWT client assertion builder (sign JWT with private key) — implemented at `src/py_identity_model/core/client_assertion.py:100` (`build_client_assertion`), produces RFC 7523 §3 claims `iss`/`sub`/`aud`/`jti`/`iat`/`exp`. Tested at `src/tests/unit/test_client_assertion.py:80` (`test_claims_ec256`), `:97`/`:114` (signature round-trips PS256/ES256), `:196` (jti uniqueness), `:1196` build (tier: unit). Note: covers private-key signing only, not HMAC.
- [AC-3] Add `client_assertion_type` and `client_assertion` to token endpoint requests — implemented at `src/py_identity_model/core/client_assertion.py:169` (`apply_private_key_jwt`) and wired into `prepare_token_request_data` (`token_client_logic.py:597`). Tested at `test_client_assertion.py:1213` (`TestApplyPrivateKeyJwt`) and end-to-end `:1390` (tier: unit + integration via `test_private_key_jwt.py:868`).
- [AC-4] Support `private_key_jwt` method (RSA/EC key pair) — implemented via `PrivateKeyJwt` model (`models.py:408`) supporting `str|bytes` PEM, algorithm default `PS256`. Tested with both EC (ES256) and RSA (PS256) keys at `test_client_assertion.py:1058`/`1068` helpers and integration at `test_private_key_jwt.py:868` (tier: unit + integration).
- [AC-6] Integrate with `ClientCredentialsTokenRequest` and grant types — `private_key_jwt` field added to ClientCredentials, AuthorizationCode, RefreshToken, Introspection, PAR, DeviceAuth, DeviceToken, TokenExchange, Revocation requests (`models.py:456,464,471,480,488,495,503,512,519`); apply hook wired into all 9 prepare functions. Tested at `test_client_assertion.py:1348` (`TestPrepareFunctionPrecedence`, parametrized over all 9 endpoints) (tier: unit) and integration for token + PAR + auth-code at `test_private_key_jwt.py` (tier: integration).
- [AC-9] Integration test: JWT assertion generation (structure, claims, signature) — `test_client_assertion.py:80-128` verifies structure/claims/signature round-trip; integration `test_private_key_jwt.py:896` (`test_client_credentials_wrong_key_rejected`) proves real signature verification at the provider (tier: unit + integration).
- [AC-10] Integration test against IdP supporting `private_key_jwt` — `test_private_key_jwt.py` runs against the node-oidc-provider fixture, which now registers a `test-private-key-jwt` client with `token_endpoint_auth_method: private_key_jwt` and a matching public JWK (`provider.js:1477`). Covers client_credentials, PAR, and authorization_code (tier: integration/e2e).
- [AC-12] Test with various signing algorithms (RS256/ES256) — covers ES256 and PS256 (`test_client_assertion.py:97,114`; parametrized unsupported-alg test `:1151`). FAPI requires PS256/ES256, both exercised (tier: unit). (RS256 not explicitly round-trip-tested but is in the supported set; PS256 substitutes as the FAPI-relevant RSA variant.)
- [AC-13] Test assertion expiration and jti uniqueness — `exp == iat + lifetime` at `test_client_assertion.py:95`; non-positive lifetime rejected `:1185`; jti uniqueness `:1196` (tier: unit).
- [AC-14] Both sync and async API paths tested — sync `test_client_assertion.py:1392`, async `:1423` (`test_async_client_credentials_sends_assertion_no_basic`); async token_client wired at `aio/token_client.py:64` (tier: unit, end-to-end with respx).
- [AC-16] Discovery of supported auth methods via `token_endpoint_auth_methods_supported` — already surfaced on the discovery model and consumed by FAPI validation (`core/fapi.py:181`); fixture advertises `private_key_jwt` (`provider.js:296` clientAuthSigningAlgValues). Pre-existing capability, integration fixture exercises it via `_is_node_oidc_fixture` gate (tier: integration). Marked PASS as the metadata path exists and is used; note no NEW discovery code in this diff.

### FAIL
- [AC-5] Support `client_secret_jwt` method (HMAC with client secret) — NOT implemented. `build_client_assertion` calls `validate_signing_algorithm`, which explicitly excludes HS256/symmetric algorithms (`jwt_signing.py:350` comment: "Symmetric (HS*) ... are intentionally excluded"). No HMAC code path, no `client_secret_jwt` model option, no test. This is an explicit spec Task and Success Criterion ("JWT client authentication working for both `private_key_jwt` and `client_secret_jwt`"). The issue title itself names both methods.
- [AC-1] Create `ClientAssertion` model (type + value) — NOT implemented as specified. The spec (mirroring Duende `ClientAssertion` with `Type`+`Value`) asks for a reusable assertion model. The diff instead adds a `PrivateKeyJwt` config dataclass (`models.py:408`) and inlines type/value into params via `apply_private_key_jwt`. No `ClientAssertion(type, value)` model exists for callers to pass a pre-built assertion. Functionally the assertion is produced, but the spec's named model deliverable is absent.
- [AC-11] Integration test against IdP supporting `client_secret_jwt` — NOT implemented (follows from AC-5 not existing). No test exists.
- [AC-15] Add JWT client authentication to FastAPI example — NOT implemented. No example file or `examples/fastapi` change in the diff demonstrates `private_key_jwt`. The only `examples/fapi_example.py:54` reference (`auth_method="private_key_jwt"`) is pre-existing and unrelated to this diff.

### PARTIAL
- [AC-7] Add to `ProtocolRequest` base class (when #88 is implemented) — the spec gates this on issue #88; #88 is not implemented, so `private_key_jwt` is added field-by-field to each request dataclass rather than on a base class. This matches the spec's conditional, so not a hard FAIL, but the deduplication the AC envisions is deferred. Acceptable given the explicit precondition.
- [AC-8] Add documentation — partial. Module/class/function docstrings are thorough (RFC citations throughout `client_assertion.py`, `models.py:408`). No user-facing documentation page or README/docs update, and no example (see AC-15). The spec lists "Documentation complete" as a Success Criterion.

### SCOPE CREEP
- `src/py_identity_model/core/jwt_signing.py:1` — extraction of `SUPPORTED_SIGNING_ALGORITHMS`/`validate_signing_algorithm` shared between JAR and client assertion is a reasonable refactor in service of AC-2/AC-4, and `jar.py` is updated to consume it. Not traceable to a specific AC but justified and low-risk; flagged for awareness only.

### Architecture Violations
- None. Business logic lives in `core/` (`client_assertion.py`, `jwt_signing.py`) and is shared by thin sync/async wrappers per the CLAUDE.md separation rule. Precedence (`private_key_jwt` > `client_secret` > public) is implemented uniformly across all prepare functions and tested at `test_client_assertion.py:1355`. Type hints present throughout. Integration fixture changes (`conftest.py`, `provider.js`) are accompanied by an actual integration test, satisfying the mandatory fixture-testing workflow rule.

### Summary
- Total ACs: 16
- Pass: 10 | Fail: 4 | Partial: 2

jamescrowley321:
## Review: review-sentinel

None.

### WARN (should fix)
None.

### INFO (acceptable risk)
- `src/tests/integration/test_private_key_jwt.py:25` and `test-fixtures/node-oidc-provider/provider.js:1495` — A static ES256 private key is checked into the repository (PEM in the test, matching public JWK in the fixture). This is a disposable test-only key, explicitly labeled as such, used only against the bundled local node-oidc-provider fixture for a `test-private-key-jwt` client. It grants no access to any real provider and the fixture itself is ephemeral. Acceptable: no production credential exposure.
- `src/py_identity_model/core/jwt_signing.py:14` — `SUPPORTED_SIGNING_ALGORITHMS` deliberately excludes `HS*` and `none`. This is the correct security posture: `validate_signing_algorithm()` is enforced in `build_client_assertion()` before signing, so a caller cannot produce a symmetric-keyed or unsigned client assertion. Prevents algorithm-confusion / `none`-alg downgrade on the assertion the library emits.
- `src/py_identity_model/core/client_assertion.py:118-130` — The assertion `aud` defaults to the request `address` (the actual endpoint URL) and may be overridden. Each assertion carries a fresh uuid4 `jti` and a bounded `iat`/`exp` (300s default), which is the correct anti-replay shape for a client-side emitter. Audience binding limits cross-endpoint replay. Acceptable.
- `src/py_identity_model/core/client_assertion.py` (and all `prepare_*` callers) — Neither the PEM private key nor the signed `client_assertion` is ever logged. `log_token_request` logs only the redacted address, `client_id`, and scope (`token_client_logic.py:33-38`); `redact_sensitive` would additionally truncate any long value. No credential leakage to logs. Acceptable.
- `src/py_identity_model/sync/token_client.py:73` / `aio/token_client.py:64` — `request_client_credentials_token` now forwards the computed `auth` (which is `None` when `private_key_jwt` is used or the client is public) instead of unconditionally sending `(client_id, client_secret)`. When `private_key_jwt` is set, no HTTP Basic header is sent and the assertion is carried in the body — verified by `test_sync_client_credentials_sends_assertion_no_basic`. Precedence (`private_key_jwt` > `client_secret` > public) is consistently applied and unit-tested across all nine client-authenticating endpoints. No fail-open path: an empty key/algorithm raises `ValueError` before any request is built. Acceptable.

### Summary
- BLOCK: 0 | WARN: 0 | INFO: 5
- Overall: PASS

jamescrowley321:
## Review: review-viper

## Review: Red Team (Viper)

### Attack Surface

This diff adds RFC 7523 `private_key_jwt` client authentication across every
client-authenticating endpoint. The security-relevant changes are:

- `core/client_assertion.py` (NEW) — builds the signed `client_assertion` JWT
  (`build_client_assertion`) and injects it into request params
  (`apply_private_key_jwt`). This is the new signing primitive that holds the
  client's private key.
- `core/jwt_signing.py` (NEW) — centralized allow-list of asymmetric signing
  algorithms (`SUPPORTED_SIGNING_ALGORITHMS`) + `validate_signing_algorithm`.
  Now governs both JAR (RFC 9101) and client assertions.
- `core/models.py` — new `PrivateKeyJwt` dataclass; `client_secret`/`scope`
  relaxed to `Optional` on `ClientCredentialsTokenRequest`; `private_key_jwt`
  field added to 8 request types.
- Auth-method selection branches in `token_client_logic`, `introspection_logic`,
  `revocation_logic`, `par_logic`, `device_auth_logic`, `token_exchange_logic`
  — each now picks `private_key_jwt` > `client_secret` > public.
- `sync/token_client.py` + `aio/token_client.py` — stop passing
  `(client_id, client_secret)` unconditionally; now honor the `auth` value
  returned by `prepare_token_request_data`.
- `test-fixtures/node-oidc-provider/provider.js` — registers a
  `private_key_jwt` test client with a checked-in public JWK (test fixture).

This is a client-side OAuth library: it *produces* assertions, it does not
*verify* them. The authorization server performs signature/`aud`/`exp`/`jti`
validation. The attacker-relevant questions are therefore: can the library be
driven to leak the private key, sign a confused/over-scoped assertion, or
silently downgrade/omit authentication?

### Findings

#### LOW — Algorithm field is caller-controlled but constrained to asymmetric allow-list (downgrade is blocked)
- **Location**: `src/py_identity_model/core/client_assertion.py:71`, `src/py_identity_model/core/jwt_signing.py:33-50`
- **Attack scenario**:
  1. A caller (or a config source feeding `PrivateKeyJwt.algorithm`) attempts
     to set `algorithm="none"` or `algorithm="HS256"` to produce an
     unsigned/symmetric assertion that an attacker who knows the client_id
     could forge.
  2. `build_client_assertion` calls `validate_signing_algorithm`, which rejects
     anything not in `SUPPORTED_SIGNING_ALGORITHMS` (asymmetric only — `none`
     and `HS*` are explicitly excluded).
  3. A `ValueError` is raised before `pyjwt.encode` runs. No assertion is
     emitted.
- **Prerequisites**: Control over the `algorithm` value. Even with it, the
  attack fails — this is the control that *prevents* the classic JWT alg-confusion
  downgrade. Reported only to document that it was checked and holds.
- **CVSS v3.1**: 0.0 (AV:N/AC:H/PR:H/UI:N/S:U/C:N/I:N/A:N) — not exploitable.
- **Remediation**: None required. The allow-list correctly excludes `none`/`HS*`.
  (Optional hardening: when used specifically for FAPI 2.0 client auth, narrow
  to `PS256`/`ES256` per FAPI 2.0 Security Profile, but the broader asymmetric
  set is acceptable for general RFC 7523 use.)

#### LOW — Assertion `aud` defaults to endpoint URL; replay across endpoints depends on AS, not library
- **Location**: `src/py_identity_model/core/client_assertion.py:118-125`, `apply_private_key_jwt`
- **Attack scenario**:
  1. The library sets the assertion `aud` to the request `address` (e.g. the
     token endpoint) by default, with optional override to the issuer.
  2. An attacker who captures a `client_assertion` in transit (requires MITM on
     an already-TLS-protected channel) tries to replay it against a different
     endpoint of the same AS, or after expiry.
  3. Replay is bounded by `jti` (uuid4, unique per build — verified by
     `test_jti_unique_across_builds`) and a 300s `exp`. The AS is responsible
     for `jti` single-use and `aud` matching; a correct AS rejects the replay.
- **Prerequisites**: TLS interception of the client→AS channel within the 300s
  window, plus an AS that does not enforce `jti`/`aud`. The library does its
  part correctly (unique `jti`, short `exp`, endpoint-scoped `aud`).
- **CVSS v3.1**: 2.6 (AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N)
- **Remediation**: None required of the library. Documentation already notes
  `aud` can be set to the issuer; for maximum interop with AS implementations
  that expect the issuer rather than the endpoint URL, callers can set
  `PrivateKeyJwt.audience` explicitly. No code change needed.

#### LOW — Relaxing `client_secret`/`scope` to Optional does not enable a silent no-auth request
- **Location**: `src/py_identity_model/core/models.py:451-456`, `src/py_identity_model/core/token_client_logic.py` (client-credentials branch), `src/py_identity_model/sync/token_client.py:78`, `src/py_identity_model/aio/token_client.py:64`
- **Attack scenario**:
  1. `ClientCredentialsTokenRequest.client_secret` is now `None`-defaulted and
     the sync/async clients no longer unconditionally send
     `(client_id, client_secret)` as HTTP Basic. Concern: a request with
     neither `client_secret` nor `private_key_jwt` could silently
     authenticate as a public client where a confidential one was intended.
  2. Inspecting the branch: with no `private_key_jwt` and no `client_secret`,
     the code sets `params["client_id"] = request.client_id` and `auth=None`
     — i.e. an explicit public-client request, which a confidential-client AS
     rejects with `invalid_client`. There is no path that drops the assertion
     while believing it was sent: `apply_private_key_jwt` always writes
     `client_assertion`/`client_assertion_type` or raises.
  3. No credential is ever sent to the wrong place: when `private_key_jwt` is
     set, `auth` is forced to `None` so no Basic header leaks the (absent)
     secret, and the assertion goes in the body. Verified by
     `TestEndToEndNoBasicHeader` (no `Authorization` header) and the
     precedence parametrized tests across all 9 endpoints.
- **Prerequisites**: Misconfiguration by the integrating application (omitting
  both auth methods). The failure mode is a rejected request, not a privilege
  gain.
- **CVSS v3.1**: 1.8 (AV:N/AC:H/PR:H/UI:N/S:U/C:N/I:L/A:N)
- **Remediation**: None required. Behavior is fail-closed at the AS. Optionally,
  the library could raise locally when a confidential grant has no auth method,
  but this is a usability nicety, not a security fix.

### Summary
- Attack surface elements: 6 (client_assertion builder, jwt_signing allow-list,
  models auth fields, 6 auth-selection branches, sync/async token clients, test
  fixture with checked-in public key)
- Findings: 0 critical, 0 high, 0 medium, 3 low
- Notes: The private key is held only in the `PrivateKeyJwt` dataclass and
  passed straight to `pyjwt.encode`; it is never logged (logging is limited to
  `client_id`/`scope`/redacted URL) and never placed in a URL or header. The
  checked-in key in the test fixture/integration test is a disposable P-256
  test key (only the public half is in `provider.js`), clearly labeled
  test-only — not a production secret exposure. Algorithm downgrade to
  `none`/`HS*` is blocked by the shared allow-list. `jti` uniqueness and short
  `exp` limit replay. No exploitable auth-bypass, privilege-escalation,
  token-confusion, or credential-leak path found in the changes.
- Overall: PASS

jamescrowley321:
## Adversarial code review — fresh-context re-review (5 layers)

Reviewed under fresh context by Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, and Viper. **Crypto is clean** — both security layers cleared assertion `aud` pinning (per-endpoint), `jti` uniqueness, `exp` bounding, the asymmetric-only alg allow-list (`none`/HMAC excluded), and precedence (fails closed, no silent downgrade). Acceptance Auditor confirmed every T57 contract item is delivered. No P0/P1 security blockers.

### Findings addressed in the latest push
- **P1 — `PrivateKeyJwt` repr leaked the PEM private key.** Default dataclass repr rendered the key (unlike sibling secret-bearing request models). Fixed: `private_key = field(repr=False)` + `private_key`/`client_assertion` added to `redact_sensitive`. Regression tests added.
- **P2 — `lifetime` ≤ 10s minted an already-expired assertion** (`exp = (now − leeway) + lifetime`, guard only rejected `≤ 0`). Now requires `lifetime > clock-skew leeway`.
- **P2 — RS256 was never exercised** (all tests PS256/ES256). Added an RS256 signing round-trip and an **async-path** `private_key_jwt` integration test (#213 asks for both sync + async and RS256 + ES256).
- **P3 — whitespace-only `client_id`/`audience`/`kid`** bypassed the empty-value guards; now rejected.
- **P3 — empty-string `client_secret`** handled inconsistently (truthiness in client-credentials vs `is not None` elsewhere); aligned on `is not None`.

### Resolved reviewer disagreement
Default assertion `aud` = the called endpoint's URL drew opposite calls (Blind/Edge flagged non-conformance for non-token endpoints; Sentinel/Viper judged it the safer, replay-minimizing choice). **Resolution: not a security bug** — per-endpoint `aud` is the conservative reading of RFC 7523 §3 and is overridable via `PrivateKeyJwt.audience`. Minor interop note only (some ASes require issuer-as-`aud` on introspection/revocation); no code change needed to merge.

### Deferred (correctly scoped out by T57; issue #213 remains open)
`client_secret_jwt`, the generic `ClientAssertion(type, value)` model, and adding JWT auth to the FastAPI example are genuinely separate from the T57 `private_key_jwt` scope — the "out of scope" calls are justified, but #213 itself should stay open until they land.

`make lint` (ruff + format + pyrefly + coverage) and unit tests pass.

PR #434, opened 2026-07-02
build(fastapi): add fastapi-identity-model package (middleware + RP router)

jamescrowley321:
## Summary

Promotes the FastAPI example into a first-class, **independently publishable** package **`fastapi-identity-model`** under `packages/`, built on the OIDF-certified `py-identity-model` core. Delivers both a resource-server middleware and a relying-party login router, plus a runnable demo.

## What's in the package (`packages/fastapi-identity-model/`)
- **`TokenValidationMiddleware`** — resource-server Bearer validation (moved from the example via `git mv`; returns **500** for unexpected server faults instead of masking as 401).
- **`build_oidc_router`** — mountable **relying-party login flow** (authorization code + PKCE, state/nonce, ID-token validation, UserInfo `sub`-check). Identity goes in the session; raw tokens only when `store_tokens=True` (documented cookie-signing caveat).
- **`Depends` helpers** (`get_current_user`, `require_scope`, `require_claim`, …), typed **`OIDCSettings`**, and an async-fixed **`TokenManager`** (uses native `aio.refresh_token`; removes the blocking sync-in-async call).
- **43 unit tests, 96% coverage** (httpx `ASGITransport`, no live network); **PEP 561 typed** (`py.typed`), pyrefly-clean.

## Repo wiring
- `uv` workspace member; `make test-fastapi` (typecheck + tests) / `make build-fastapi`; new **`fastapi-tests` CI job**.
- **Independent release** — `.github/workflows/release-fastapi.yml` (manual dispatch or `fastapi-identity-model-v*` tag, **PyPI trusted publishing / OIDC**). Own version (`0.1.0`) + `CHANGELOG`. **Leaves the root `semantic-release` pipeline untouched**: package-only changes use non-releasing commit types scoped `(fastapi)` so they never bump `py-identity-model`.
- `examples/fastapi` rewired into a runnable demo (browser login → session `/me`, plus Bearer-protected `/api/*`); `examples/descope` repointed at the package.

## Verification
- `make lint` ✅ (ruff, ruff-format, pyrefly, coverage)
- `make test-fastapi` ✅ — 43 passed, 96.47% coverage, pyrefly 0 errors
- `uv build --package fastapi-identity-model` ✅ — wheel + sdist, `twine check` PASSED, ships `py.typed`

## ⚠️ One-time owner step before first publish
Register a **PyPI Trusted Publisher** for project `fastapi-identity-model` pointing at this repo + workflow `release-fastapi.yml` + environment `pypi`. No API token is stored. Until then, `release-fastapi` builds/validates but the publish step will fail by design.


jamescrowley321:
## Adversarial code review — fresh-context re-review (5 layers)

Reviewed under fresh context by Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, and Viper. The **token-validation crypto is sound** (delegated to the certified core; all layers cleared signature/issuer/audience/nonce/PKCE/state). The issues were in the RP router edge handling, the TokenManager, the middleware token-type/audience defaults, and the example wiring — all addressed in the latest push.

### P0 — CI (fixed)
The four example Dockerfiles copied the workspace root pyproject (which lists `packages/fastapi-identity-model` as a member) but never the member directory, so `uv sync --all-packages` failed with *"not a workspace member"* — the `example-tests` failure. Reproduced locally; fixed by copying the package into each build context. (The pyproject metadata itself was consistent — this was purely the missing `COPY`.)

### P1 (fixed)
- **Fail-open login:** a token response without an `id_token` still created a session (nonce check skipped, `sub=None`). Now rejected.
- **Shared session key:** flow state and identity shared `session["oidc"]`, so opening `/auth/login` read as authenticated and leaked the `code_verifier`. Flow state moved to a separate single-use key.
- **Insecure example:** published fallback `SESSION_SECRET` + `store_tokens=True` on an unencrypted cookie (Viper rated this a full auth-bypass chain if copied to prod). Now fails closed on missing secret, defaults `store_tokens=False`, sets cookie flags.
- **Middleware ID-vs-access token:** with `audience` defaulting to `client_id`, an ID token validated as an access token. Now rejects ID-token-only claims (`nonce`/`at_hash`/`c_hash`) and requires a non-empty `audience`.

### P2/P3 (fixed)
Middleware surfaces discovery/JWKS faults as **503** (not 401), passes CORS preflight, matches excluded subpaths, respects explicit `excluded_paths=[]`. TokenManager refreshes under an `asyncio.Lock` and treats `expires_in=0` as expired. Config: `redirect_uri` optional for resource-server-only, empty required fields rejected. `require_scope` fails closed on a non-string/list claim. `POST /auth/logout` (CSRF-safe), `exp` added to ID-token require, `issuer` guarded, malformed callbacks → 400, SessionMiddleware presence validated. `fastapi` floor bumped to `>=0.115.0` (0.100 can't satisfy `starlette>=0.49.1`). Release workflow: restored `contents: read`, tests before publish, publish tag-only.

### Verified false positives (dropped)
- *"pyrefly typecheck is a silent no-op"* — tested pyrefly 1.1.1: an explicit CLI path **is** checked despite `project_excludes`.
- *"async tests never run"* — root sets `asyncio_mode = "auto"`.
- *"workspace source references a non-member"* — metadata is consistent; the failure was the Docker `COPY` gap only.

### Deferred (core-library scope, out of this package PR)
- The middleware can't fully distinguish transient network faults from bad tokens because the **core library** re-raises discovery/JWKS fetch failures as `TokenValidationException` (not `NetworkException`). Mitigated here (503 on `NetworkException`); a complete fix belongs in `py_identity_model` core.
- `to_principal` stringifies claim values (`True`→`"True"`), so `require_claim("email_verified","true")` never matches — core `identity.py`, fail-closed, tracked separately.
- Release-scope isolation is convention-only (a `feat(fastapi)` commit could bump root); enforcing it means changing root semantic-release config — deliberately not done in this package PR.

`make test-fastapi` (58 passed, 95% coverage, pyrefly clean) and `make lint` pass.

PR #447, opened 2026-07-21
feat(integration): wire Keycloak fixture into integration harness

jamescrowley321:
## Summary
Wires the KC.1 Keycloak fixture into the integration harness. No library/core changes (that is KC.3 scope).

- **`.env.keycloak`** (new, tracked) — mirrors the `.env.node-oidc` contract; discovery + JWKS URLs, `py-identity-model-client` client-credentials creds, `test-auth-code` / `test-pkce-public` client vars, `TEST_SCOPE=openid`, `TEST_REQUIRE_HTTPS=false`. Opaque vars omitted (Keycloak emits JWTs → `TEST_OPAQUE_*` tests skip cleanly).
- **`.gitignore`** — add `!.env.keycloak` after the `!.env.node-oidc` allow-rule so the tracked fixture env escapes the `*.env*` blanket ignore.
- **`Makefile`** — add `test-integration-keycloak` target, an exact mirror of `test-integration-node-oidc` (`docker compose ... up -d --build --wait` → `pytest -m integration --env-file=.env.keycloak` → teardown).
- **`provider_matrix.py`** — no code change needed; `find_env_files()` globs `.env*` (skipping `.example`/`.bak`), so the tracked `.env.keycloak` is auto-discovered. Confirmed live: a `keycloak` column appears in `make provider-matrix`.

Story: KC.2 (Keycloak provider wiring). AC source: issue #244.

> **Recommend:** base this on top of the KC.1 fixture PR #446 (this PR is stacked on `feat/keycloak-fixture`). Merge #446 first, then this.

## ACs covered (KC.2 wiring)
- AC — env file for the harness → `.env.keycloak`
- AC — `make test-integration-keycloak` target (mirror node-oidc) → `Makefile`
- AC — `provider_matrix.py` discovers Keycloak → confirmed live (`keycloak` column, 12/13 caps ✓, DPoP ✗ as expected)

The full capability-gated suite going green vs Keycloak (36 currently-red tests: httpx pool intermittent connection-refused, hardcoded node-oidc tokens, dev-login helper, DPoP capability-skip) is **deferred to KC.3** by design — this PR is wiring only.

## Review Findings Addressed
Reviewers (KC.x override): Blind Hunter + Edge Case Hunter + Acceptance Auditor.
- **P0: 0** (Blind 0 MUST FIX / Acceptance 0 FAIL / Edge 0 critical).
[...]

PR #448, opened 2026-07-21
test(keycloak): full capability-gated integration suite green + x5t#S256 core fix

jamescrowley321:
## Summary

Makes the full integration suite run **green against live Keycloak** (KC.3), fixing one genuine `core/` serialization gap plus the test-infra/fixture wiring that live Keycloak surfaced. Capability-gated skips remain clean.

**Core library fix (sync+async shared model):**
- `fix(jwks)` — `JsonWebKey.as_dict()` now emits the RFC 7517 member name `x5t#S256` (was the Python attr name `x5t_s256`), so it round-trips with `to_json()`/`jwks_from_dict()`. Keycloak JWKS carries `x5t#S256`; node-oidc never did, so it was never exercised. Added a unit round-trip test in `test_parsers.py`.

**Test-infra / fixture (Keycloak-specific, node-oidc kept byte-identical):**
- `test(keycloak)` — key all 5 cross-worker session caches/locks by a `provider_slug` derived from the env file, killing cross-provider cache poisoning (`/var/tmp/pytest-of-*` reuse of stale node-oidc issuer/kid/tokens).
- `test(keycloak)` — generalize the auth-code login helper to parse the returned HTML `<form>` (action + field names) for Keycloak while preserving the node-oidc devInteractions path (branch on `/interaction/`).
- `test(keycloak)` — `_AuthFlowClient` sends Keycloak's `Secure;SameSite=None` auth-session cookies over http (httpx 0.28.1 refuses to and can't be forced).
- `test(keycloak)` — gate `private_key_jwt` tests to the node-oidc fixture (DEC-007 option B); skip KC token-exchange rejection cleanly.
- `build(keycloak)` — realm-export: enable device grant + standard token exchange on the client, set test-user password to `test` (matches `AuthCodeFlowConfig` default), grant `offline_access`/`uma_authorization` roles.

Refs KC.3 (Keycloak provider integration epic)

## Base

Stacked on **#447** (`feat/keycloak-provider-wiring`), which is stacked on **#446** (`feat/keycloak-fixture`). Recommend merging #446 → #447 → this, bottom-up.

## Review Findings Addressed

Reviewers (KC.x): Blind Hunter + Edge Case Hunter + Acceptance Auditor.
- **Acceptance:** 4/4 ACs PASS — full capability-gated suite green vs Keycloak; only prod-code change is the `x5t#S256` serialization fix with a unit test.
- **P0:** 0.
- **P1:** 3 fixed (double-unescape of form action; identifier-field fallback with loud pytest.fail; nested-`<form>` depth counter), 1 skipped-with-rationale (`_AuthFlowClient` cookie replay — single-host localhost fixtures only).
- **Delta re-review (iter-2):** 0 MUST / 0 SHOULD, no regressions.

## Test plan
- [x] Unit tests pass (1223 passed, 95.5% cov)
- [x] Integration tests pass — Keycloak 93 passed / 29 clean skips / 0 fail; node-oidc regression 117 passed / 5 skip / 0 fail
- [x] Lint passes (ruff + ruff format + pyrefly + coverage gate)
- [ ] CI passes

### DIFF
--- a/.env.keycloak
+++ b/.env.keycloak
@@ -1,0 +1,12 @@
+TEST_DISCO_ADDRESS=http://localhost:8080/realms/py-identity-model/.well-known/openid-configuration
+TEST_JWKS_ADDRESS=http://localhost:8080/realms/py-identity-model/protocol/openid-connect/certs
+TEST_CLIENT_ID=py-identity-model-client
+TEST_CLIENT_SECRET=py-identity-model-client-secret
+TEST_SCOPE=openid
+TEST_AUDIENCE=
+TEST_REQUIRE_HTTPS=false
+TEST_AUTH_CODE_CLIENT_ID=test-auth-code
+TEST_AUTH_CODE_CLIENT_SECRET=test-auth-code-secret
+TEST_AUTH_CODE_REDIRECT_URI=http://localhost:8080/callback
+TEST_PKCE_PUBLIC_CLIENT_ID=test-pkce-public
+TEST_PKCE_PUBLIC_REDIRECT_URI=http://localhost:8080/callback
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,7 @@ dmypy.json
 *.env*
 # Allow test fixture env files (no secrets — local Docker only)
 !.env.node-oidc
+!.env.keycloak
 *.crt
 *.key
 *.pfx
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,15 @@ test-integration-node-oidc: ## Run integration tests against node-oidc-provider
 		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
 
+.PHONY: test-integration-keycloak
+test-integration-keycloak: ## Run integration tests against Keycloak
+	@echo "Starting Keycloak fixture..."
+	docker compose -f test-fixtures/keycloak/docker-compose.yml up -d --build --wait
+	@echo "Running integration tests against Keycloak..."
+	uv run pytest src/tests -m integration --env-file=.env.keycloak -v || \
+		(docker compose -f test-fixtures/keycloak/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/keycloak/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/src/py_identity_model/core/models.py
+++ b/src/py_identity_model/core/models.py
@@ -388,9 +388,25 @@ class JsonWebKey:
         return urlsafe_b64decode(input_str + padding)
 
     def as_dict(self):
-        """Convert the JWK to a dictionary with all available properties"""
-        # Add all non-None properties to the dictionary
-        return {key: value for key, value in self.__dict__.items() if value is not None}
+        """Convert the JWK to a dictionary with all available properties.
+
+        Python attribute names are mapped back to their RFC 7517 JWK member
+        names (e.g. ``x5t_s256`` -> ``x5t#S256``) so the result round-trips
+        through :meth:`from_json` and :func:`jwks_from_dict`. Providers such as
+        Keycloak emit ``x5t#S256`` in their JWKS; without this mapping the
+        dictionary form used ``x5t_s256`` and silently dropped the thumbprint
+        on re-parse.
+        """
+        data = {}
+        for key, value in self.__dict__.items():
+            if value is None:
+                continue
+            # Map Python field names back to JWK parameter names
+            if key == "x5t_s256":
+                data[JsonWebKeyParameterNames.X5T_S256.value] = value
+            else:
+                data[key] = value
+        return data
 
 
 # ============================================================================
--- a/src/tests/integration/conftest.py
+++ b/src/tests/integration/conftest.py
@@ -8,11 +8,13 @@ Includes:
 
 from contextlib import suppress
 from dataclasses import dataclass
+from html import unescape
+from html.parser import HTMLParser
 import json
 import secrets
 import time
 from types import MappingProxyType
-from urllib.parse import urlencode, urlparse
+from urllib.parse import urlencode, urljoin, urlparse
 
 from filelock import FileLock
 import httpx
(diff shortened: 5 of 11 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/tests/integration/conftest.py:257`] `_parse_login_form` calls `unescape(parser.action)`, but `html.parser.HTMLParser` already unescapes character references in attribute values (`parse_starttag` runs `attrvalue = unescape(attrvalue)`). So the action URL is unescaped twice while the hidden input values (`parser.inputs`) are unescaped exactly once. This is inconsistent handling of two things pulled from the same parser, and a double-unescape can corrupt an action URL that legitimately contains a double-encoded sequence (e.g. `&amp;amp;` decodes to `&` instead of `&amp;`), sending a malformed POST target.
- [`src/tests/integration/conftest.py:299`] `_AuthFlowClient.send` replays every captured cookie on every outgoing request unconditionally — it ignores cookie Domain/Path/host entirely. The store keeps `name -> value` with no host key, so if the auth-code flow ever redirects to a different host (e.g. a redirect_uri or IdP on another origin), Secure auth-session cookies (`AUTH_SESSION_ID`, `KC_RESTART`) are transmitted to that unrelated host. The whole point of the added class is to bypass the policy that normally prevents this, so nothing else guards it.

### NITPICK
- [`src/tests/integration/conftest.py:308`] `_iter_set_cookies` collapses cookies of the same name from different paths/domains into a single key in `_flow_cookies`; a later `Set-Cookie` silently overwrites an earlier one. Fine for the current single-host fixtures, but a latent correctness gap given the "share one code path for both providers" goal.
- [`src/tests/integration/conftest.py:76`] `provider_slug` falls back to `"default"` for any env file not matching the `.env.<name>` shape, so two differently-configured-but-misnamed providers would still collide on the same cache files — the exact cross-provider cache poisoning the fixture claims to fix.
- [`src/py_identity_model/core/models.py:27`] `as_dict` now references `JsonWebKeyParameterNames.X5T_S256.value`; the diff does not add an import for `JsonWebKeyParameterNames`. If the symbol is not already imported in `models.py`, this raises `NameError` on every `as_dict()` call. Worth confirming the import is present.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `src/tests/integration/conftest.py:342` | HTML login form names its identifier field neither `username` nor `login` | `data[next(iter(field_names - {"password"}))] = config.login_hint` (or fail with a clear message) | [DEGRADED] no identifier sent; POST fails, callback never reached, opaque downstream `pytest.fail` |
| `src/tests/integration/conftest.py:676` | Provider page has a password form nested inside another form (or two password forms) | reset `_current_*` only when `not self._in_form` before overwriting | [WRONG] inner `<form>` reset drops outer form's action/inputs, wrong POST target parsed |

### Summary
- Unhandled paths found: 2

jamescrowley321:
## Review: Acceptance Auditor

### PASS
- [AC-2] `x5t#S256` (RFC 7517) serialization round-trips correctly — implemented at `src/py_identity_model/core/models.py:400` (`JsonWebKey.as_dict` now maps `x5t_s256` → `x5t#S256` via `JsonWebKeyParameterNames.X5T_S256.value`, mirroring the existing `to_json`/`from_json` mapping at lines 356/323). Covered by a NEW unit test at `src/tests/unit/test_parsers.py:477` (`test_x5t_s256_round_trips_via_as_dict`, tier: unit) which asserts `as_dict()["x5t#S256"]`, absence of `x5t_s256`, and full round-trip through both `jwks_from_dict` and `JsonWebKey.from_json`. I executed the test in isolation — it PASSES and would fail if the mapping were removed (it explicitly asserts `"x5t_s256" not in as_dict`). Integration coverage additionally added at `src/tests/integration/test_json_web_key.py:36,128,139` (tier: integration).

- [AC-3] No cross-provider regression from shared `conftest.py` changes — PASS. Cache/lock keying is namespaced by `provider_slug` (`src/tests/integration/conftest.py:96-116`) derived from the `--env-file` name (`.env.node-oidc` → `node-oidc`, `.env.keycloak` → `keycloak`), applied consistently to all five session caches (test_config, discovery_document, raw_discovery, jwks_response, client_credentials_token, opaque_access_token) at lines 84,101-103,120-122,138-141,163-166,188-191. The node-oidc login path is preserved: `_submit_login` (`conftest.py:775`) dispatches on `"/interaction/" in interaction_url` and posts the byte-for-byte identical `{prompt/login/password}` payload for node-oidc, only branching to HTML-form parsing otherwise. `AuthCodeFlowConfig` defaults (`login_hint="test-user"`, `login_password="test"`, lines 811-812) are unchanged; the Keycloak realm fixture was adjusted to match (password `test`) rather than altering the shared default. Verification of AC-3 is the existing node-oidc suite (`make test-integration-node-oidc`); no live run was reproducible in this environment, but the changes are structurally transparent to node-oidc.

- [AC-4] Capability-gating, not masking — PASS. Skips are gated on positive capability detection or on explicit provider-unsupported signals, not blanket suppression:
  - Device auth: `test_device_token_exchange.py:96-98,116-118,148-150` skip only when `device_authorization`/`device_authorization_endpoint` are absent from `provider_capabilities`; a real request failure still asserts (`assert response.is_successful is True`).
  - Token exchange: skip only on the explicit `_UNSUPPORTED_GRANT_SIGNALS` match (`test_device_token_exchange.py:82-89`, `_provider_lacks_token_exchange`); any other failure fails the test.
  - private_key_jwt: `test_private_key_jwt.py:99-105` skips only when the provider is not the node-oidc host-root fixture that registers the static test client — the tightened `_is_node_oidc_fixture` check (hostname + host-root path) correctly excludes Keycloak (served under `/realms/<realm>`) without masking a library defect, since the static key genuinely isn't registered there.
  - Opaque-token/introspection tests skip via absent `TEST_OPAQUE_CLIENT_ID` (documented; Keycloak emits JWTs).

- [AC-1] Full integration suite runs green vs live Keycloak — PASS (with caveat). All machinery required to convert Keycloak failures into passes or clean capability-gated skips is present: (a) generic HTML login-form parser `_LoginFormParser`/`_parse_login_form` (`conftest.py:200-264`) handling Keycloak's server-rendered form, hidden fields (`credentialId`), and HTML-entity-unescaped action URLs; (b) `_AuthFlowClient` (`conftest.py:276-311`) that echoes back Keycloak's `Secure; SameSite=None` auth-session cookies over plain http, which stdlib httpx policy would otherwise drop; (c) provider-isolated caches (AC-3); (d) realm fixture wiring device grant + standard token exchange (`realm-export.json:27-31`) and the test-user roles/password (`realm-export.json:88,91`); (e) the capability skips of AC-4 for genuinely-absent features. CAVEAT: a live green run against a running Keycloak could not be independently reproduced in this review environment; PASS is on completeness/correctness of the delivered fixtures, conftest, and tests, which per the charter constitute the integration coverage.

### FAIL
- (none)

### PARTIAL
- (none)

### SCOPE CREEP
- `test-fixtures/keycloak/realm-export.json:27-31` — enabling `oauth2.device.authorization.grant.enabled` and `standard.token.exchange.enabled` on `py-identity-model-client` makes the device-grant and token-exchange integration tests actively RUN (assert success) rather than skip. This exceeds the minimal "make the suite green" bar but is squarely within the charter's "full capability-gated integration suite" intent and increases real coverage; noted, not a defect. The only production-code change in the diff is the `x5t#S256` fix in `core/models.py`, matching the task's stated expectation — no unexpected production scope creep.

### Summary
- Total ACs: 4

Issue #442, opened 2026-07-08
cert(logout): Back-Channel Logout RP — validate_logout_token (OIDC Back-Channel Logout 1.0)


jamescrowley321:
## Goal

Add **Back-Channel Logout RP** support (`oidcc-client-backchannel-logout-certification-test-plan`): validate an OP-sent **Logout Token** per [OpenID Connect Back-Channel Logout 1.0 §2.4](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation). Listed as a target ("nice-to-have") in tracker #242 but has no issue of its own.

## Why this fits the library

Back-channel logout is a **server-to-server JWT** — validating it is squarely the library's core competency (signature + claim checks). This is the highest-fit of the remaining logout profiles.

## Scope — core library: `validate_logout_token()`

Validate a Logout Token JWT:
- Standard JWT checks: signature (JWKS), `iss`, `aud`, `iat`, `exp` (if present).
- MUST contain `iss`, `aud`, `iat`, `jti`, `events`.
- MUST contain `sub` and/or `sid` (at least one).
- `events` MUST contain the member `http://schemas.openid.net/event/backchannel-logout` (value `{}`).
- MUST **reject** a `nonce` claim.
- SHOULD honor `typ: logout+jwt`.
- Add discovery fields: `backchannel_logout_supported`, `backchannel_logout_session_supported`.

Sync + async parity; unit tests for each reject rule (missing `events`, wrong `events`, expired, no `sub`/`sid`, `nonce` present); integration test; example.

## Middleware split (open branch `feat/fastapi-identity-model-package`)

The core library validates the token. The **`fastapi-identity-model`** package owns the HTTP side: expose the `backchannel_logout_uri` POST receiver, call `validate_logout_token`, then terminate the session keyed by `sid`/`sub`. Track that endpoint work against the middleware package.

## Conformance tests targeted

LOGOUT-004…007, LOGOUT-009, LOGOUT-010 (per planning epic `epic-0e-spec-logout.md`, story S.10b).

## References

- Tracking: #242 · RP-Initiated Logout sibling: #214 · Planning: identity-stack-planning `epic-0e-spec-logout.md` (S.10b)
- [Back-Channel Logout 1.0 §2.4 — Logout Token Validation](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation)

jamescrowley321:
[...]

PR #450, opened 2026-07-21
feat(logout): back-channel logout RP — validate_logout_token

jamescrowley321:
## Summary

Implements the **Back-Channel Logout RP** profile (OpenID Connect Back-Channel Logout 1.0 §2.4/§2.5) for the next certification round.

- **`core/logout_logic.py`** — pure `validate_logout_token_claims(claims)` enforcing the spec reject rules: required `iss`/`aud`/`iat`/`jti`/`events`; reject `nonce`; require ≥1 of `sub`/`sid`; `events` must be a JSON object containing the `http://schemas.openid.net/event/backchannel-logout` member. Reuses existing `oidc_constants` (`Events.BACK_CHANNEL_LOGOUT`, `BackChannelLogoutRequest.LOGOUT_TOKEN`).
- **`sync/logout.py` + `aio/logout.py`** — thin `validate_logout_token(...)` wrappers that first run the standard `validate_token` (JWKS signature + `iss` + `aud` + `exp`-if-present) then the shared pure claim check. Sync/async parity guaranteed by sharing the pure function.
- **Discovery model** — added `backchannel_logout_supported` / `backchannel_logout_session_supported` (`bool | None`) to `DiscoveryDocumentResponse` + `_guarded_fields` + `build_discovery_response` mapping.
- **Exceptions/exports** — `LogoutTokenValidationException(TokenValidationException)`; `validate_logout_token` exported from sync/aio/top-level `__init__`.
- **Conformance harness** — `POST /backchannel-logout` receiver + `configs/backchannel-logout-rp.json`.

Refs #442

## Spec IDs / ACs covered
- **LOGOUT-004** accept — valid signed logout token returns claims (full respx disco+JWKS flow)
- **LOGOUT-005** missing `events` → `LogoutTokenValidationException`
- **LOGOUT-006** wrong `events` member → rejected
- **LOGOUT-007** expired (`exp` past) → `TokenExpiredException` from `validate_token`
- **LOGOUT-009** no `sub` and no `sid` → rejected
- **LOGOUT-010** `nonce` present → rejected
- Discovery flag ACs — `backchannel_logout_supported` / `_session_supported` populate from JSON, absent → `None`
- Sync + async parity asserted for every case (front-channel LOGOUT-008 + session mgmt are out of scope / dropped)

## Review Findings Addressed
- Reviewers: Blind Hunter + Edge Case Hunter + Acceptance Auditor + Sentinel + Viper (security-sensitive JWT tier). **0 P0.**
- Acceptance Auditor 6/6 PASS; Sentinel PASS (0 BLOCK); Viper PASS (0 crit/high/med).
- **2 P1 fixed** (commit `337a463`, `conformance/app.py`):
  1. Edge [WRONG] / Blind SHOULD — malformed non-JWT `logout_token` raised PyJWT `DecodeError` (a `PyJWTError`, disjoint from `PyIdentityModelException`) → previously uncaught HTTP 500. Now caught → 400 per §2.5.
  2. Edge [WRONG] / Blind SHOULD — TOCTOU on `_last_flow_context`; snapshot once before reading fields.
- **P1 skipped w/ rationale:** committed `client_secret` placeholder (standard OIDF cert-config convention, not a real credential); `/authorize` write-lock (cert harness driven serially + snapshot removes the splice hazard).
- **P2 skipped:** Viper 3 LOW (jti replay / aud-None / harness `require_https`) + Blind 2 nitpicks — hardening/cosmetic, not exploitable against the shipped lib.
- Delta re-review (Edge + Blind on the fix diff): 0 MUST / 0 SHOULD, fixes correct & complete, no regressions.

## Test plan
- [x] Unit tests pass — `make test-unit`: 1247 passed, 95.56% cov (sync/logout.py 100%); LOGOUT-004..010 + discovery-field tests, sync + aio
- [x] Integration tests — `[skip-integration-tests]` justified: pure lib logic + harness + unit tests only, no conftest/fixture edits
- [x] Lint passes — `make lint` (ruff / format / pyrefly / pytest-80%)
- [ ] CI passes

## Recommend
Base off #449 (`feat/provider-matrix-logout-columns`, still OPEN) per the base-branch chain. Owner merges bottom-up.

### DIFF
--- a/conformance/app.py
+++ b/conformance/app.py
@@ -19,6 +19,7 @@ from urllib.parse import urlencode
 
 from fastapi import FastAPI, Query, Request
 from fastapi.responses import HTMLResponse, JSONResponse, RedirectResponse
+from jwt.exceptions import PyJWTError
 from starlette.concurrency import run_in_threadpool
 
 from py_identity_model import (
@@ -858,7 +859,12 @@ def _receive_backchannel_logout(logout_token: str | None) -> JSONResponse:
             content={"error": "invalid_request", "detail": "missing logout_token"},
         )
 
-    if _last_flow_context is None:
+    # Snapshot the process-global flow context once. A concurrent ``/authorize``
+    # can replace ``_last_flow_context`` between the None-check and the field
+    # reads; binding the reference once ensures issuer/client_id/disco all come
+    # from a single, self-consistent flow rather than a spliced mix.
+    flow_context = _last_flow_context
+    if flow_context is None:
         logger.error("REJECTED: back-channel logout received before any auth flow")
         return JSONResponse(
             status_code=400,
@@ -870,11 +876,11 @@ def _receive_backchannel_logout(logout_token: str | None) -> JSONResponse:
             logout_token,
             TokenValidationConfig(
                 perform_disco=True,
-                audience=_last_flow_context["client_id"],
-                issuer=_last_flow_context["issuer"],
+                audience=flow_context["client_id"],
+                issuer=flow_context["issuer"],
                 require_https=False,
             ),
-            disco_doc_address=_last_flow_context["disco_address"],
+            disco_doc_address=flow_context["disco_address"],
         )
     except LogoutTokenValidationException as exc:
         logger.error("REJECTED: logout token failed logout rules: %s", exc.message)
@@ -893,6 +899,16 @@ def _receive_backchannel_logout(logout_token: str | None) -> JSONResponse:
         return JSONResponse(
             status_code=400,
             content={"error": "invalid_logout_token", "detail": str(exc)},
+        )
+    except PyJWTError as exc:
+        # A non-empty but malformed ``logout_token`` (not a well-formed JWT)
+        # trips PyJWT's header decode inside ``validate_token`` and raises a
+        # ``PyJWTError`` — which is NOT a ``PyIdentityModelException``. Spec
+        # §2.5 mandates 400 (not an uncaught 500) for an invalid token.
+        logger.error("REJECTED: logout token is not a well-formed JWT: %s", exc)
+        return JSONResponse(
+            status_code=400,
+            content={"error": "invalid_logout_token", "detail": "malformed token"},
         )
 
     logger.info(
--- a/conformance/app.py
+++ b/conformance/app.py
@@ -37,10 +37,12 @@ from py_identity_model import (
     parse_discovery_url,
     request_authorization_code_token,
     validate_authorize_callback_state,
+    validate_logout_token,
     validate_token,
 )
 from py_identity_model.exceptions import (
     ConfigurationException,
+    LogoutTokenValidationException,
     PyIdentityModelException,
     TokenValidationException,
 )
@@ -210,6 +212,12 @@ test_results: dict[str, AuthSession] = {}
 
 # Shared HTTP client that skips TLS verification (conformance suite uses self-signed certs)
 _http_client: HTTPClient | None = None
+
+# Most recent auth-flow context (issuer / client_id / discovery address), captured
+# in /authorize. The OP posts a Back-Channel Logout Token to a fixed URI that
+# carries no flow parameters, so the receiver recovers the issuer (for signature
+# + iss validation) and client_id (audience) from the last flow it drove.
+_last_flow_context: dict[str, str] | None = None
 
 
 def _get_http_client() -> HTTPClient:
@@ -463,6 +471,16 @@ def authorize(
     sessions[state] = session
     if test_id:
         session.result["test_id"] = test_id
+
+    # Record the flow context so the Back-Channel Logout receiver can validate a
+    # Logout Token the OP later posts to /backchannel-logout (no flow params ride
+    # along on that request).
+    global _last_flow_context
+    _last_flow_context = {
+        "issuer": issuer,
+        "client_id": client_id,
+        "disco_address": disco_endpoint.url,
+    }
 
     logger.info(
         "ACCEPTED: discovery issuer '%s' matches expected; redirecting to OP", issuer
@@ -823,6 +841,80 @@ async def callback_post(request: Request) -> HTMLResponse | JSONResponse:
     params = urlencode(list(form_data.multi_items()))
     callback_url = f"{RP_BASE_URL}/callback?{params}"
     return await run_in_threadpool(_handle_callback, callback_url)
+
+
+def _receive_backchannel_logout(logout_token: str | None) -> JSONResponse:
+    """Validate a Back-Channel Logout Token posted by the OP (sync worker).
+
+    OpenID Connect Back-Channel Logout 1.0 §2.4/§2.5: the OP POSTs a
+    ``logout_token`` to the RP's ``backchannel_logout_uri``. The RP validates it
+    and responds 200 on success or 400 on failure. Validation is delegated to
+    ``validate_logout_token`` so the harness exercises the library's public API.
+    """
+    if not logout_token:
+        logger.error("REJECTED: back-channel logout request missing logout_token")
+        return JSONResponse(
+            status_code=400,
+            content={"error": "invalid_request", "detail": "missing logout_token"},
+        )
+
+    if _last_flow_context is None:
+        logger.error("REJECTED: back-channel logout received before any auth flow")
+        return JSONResponse(
+            status_code=400,
+            content={"error": "no_flow_context", "detail": "no prior auth flow"},
+        )
+
+    try:
+        claims = validate_logout_token(
+            logout_token,
+            TokenValidationConfig(
+                perform_disco=True,
+                audience=_last_flow_context["client_id"],
+                issuer=_last_flow_context["issuer"],
+                require_https=False,
+            ),
+            disco_doc_address=_last_flow_context["disco_address"],
+        )
+    except LogoutTokenValidationException as exc:
+        logger.error("REJECTED: logout token failed logout rules: %s", exc.message)
+        return JSONResponse(
+            status_code=400,
+            content={"error": "invalid_logout_token", "detail": exc.message},
+        )
+    except TokenValidationException as exc:
+        logger.error("REJECTED: logout token failed JWT validation: %s", exc.message)
+        return JSONResponse(
+            status_code=400,
+            content={"error": "invalid_logout_token", "detail": exc.message},
+        )
+    except PyIdentityModelException as exc:
+        logger.error("REJECTED: logout token validation error: %s", exc)
+        return JSONResponse(
+            status_code=400,
+            content={"error": "invalid_logout_token", "detail": str(exc)},
+        )
+
+    logger.info(
+        "ACCEPTED: back-channel logout token validated (sub=%s, sid=%s)",
+        claims.get("sub"),
+        claims.get("sid"),
+    )
+    return JSONResponse(content={"status": "ok"})
+
+
+@app.post("/backchannel-logout", response_model=None)
+async def backchannel_logout(request: Request) -> JSONResponse:
+    """Back-Channel Logout receiver (OpenID Connect Back-Channel Logout 1.0 §2.5).
+
+    Async to support ``await request.form()``; the blocking validation is
+    offloaded to a threadpool so the library owns its (thread-local) HTTP client.
+    """
+    form_data = await request.form()
+    logout_token = form_data.get("logout_token")
+    if not isinstance(logout_token, str):
+        logout_token = None
+    return await run_in_threadpool(_receive_backchannel_logout, logout_token)
 
 
 @app.get("/results/{test_id}")
--- a/conformance/configs/backchannel-logout-rp.json
+++ b/conformance/configs/backchannel-logout-rp.json
@@ -1,0 +1,14 @@
+{
+  "plan_name": "oidcc-client-backchannel-logout-certification-test-plan",
+  "variant": {
+    "client_registration": "static_client",
+    "request_type": "plain_http_request"
+  },
+  "alias": "py-identity-model-backchannel-logout-rp",
+  "client": {
+    "client_id": "conformance-rp",
+    "client_secret": "conformance-rp-secret",
+    "backchannel_logout_uri": "http://localhost:8888/backchannel-logout",
+    "backchannel_logout_session_required": true
+  }
+}
--- a/src/py_identity_model/__init__.py
+++ b/src/py_identity_model/__init__.py
@@ -15,6 +15,7 @@ from .exceptions import (
     InvalidAudienceException,
     InvalidIssuerException,
     JwksException,
+    LogoutTokenValidationException,
     NetworkException,
     PyIdentityModelException,
     SignatureVerificationException,
@@ -103,6 +104,7 @@ from .sync import (
     validate_fapi_authorization_request,
     validate_fapi_client_config,
     validate_fapi_discovery,
+    validate_logout_token,
     validate_token,
 )
 
@@ -155,6 +157,7 @@ __all__ = [
     "JwksException",
     "JwksRequest",
     "JwksResponse",
+    "LogoutTokenValidationException",
     "NetworkException",
     # Client authentication (private_key_jwt)
     "PrivateKeyJwt",
@@ -219,5 +222,6 @@ __all__ = [
     "validate_fapi_authorization_request",
     "validate_fapi_client_config",
     "validate_fapi_discovery",
+    "validate_logout_token",
     "validate_token",
 ]
--- a/src/py_identity_model/aio/__init__.py
+++ b/src/py_identity_model/aio/__init__.py
@@ -87,6 +87,7 @@ from .jwks import (
     get_jwks,
     jwks_from_dict,
 )
+from .logout import validate_logout_token
 from .managed_client import AsyncHTTPClient
 from .par import (
     PushedAuthorizationRequest,
@@ -209,5 +210,6 @@ __all__ = [
     "validate_fapi_authorization_request",
     "validate_fapi_client_config",
     "validate_fapi_discovery",
+    "validate_logout_token",
     "validate_token",
 ]
--- a/src/py_identity_model/aio/logout.py
+++ b/src/py_identity_model/aio/logout.py
@@ -1,0 +1,64 @@
+"""Asynchronous Back-Channel Logout token validation.
+
+Async counterpart of :mod:`py_identity_model.sync.logout`. Awaits the standard
+async ``validate_token`` then applies the same pure Logout-Token claim rules
+from ``core.logout_logic`` — sync/async parity is guaranteed by sharing that
+single pure validation function (OpenID Connect Back-Channel Logout 1.0 §2.4).
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from ..core.logout_logic import validate_logout_token_claims
+from .token_validation import validate_token
+
+
+if TYPE_CHECKING:
+    from ..core.models import TokenValidationConfig
+    from .managed_client import AsyncHTTPClient
+
+
+async def validate_logout_token(
+    logout_token: str,
+    token_validation_config: TokenValidationConfig,
+    disco_doc_address: str | None = None,
+    http_client: AsyncHTTPClient | None = None,
+) -> dict:
+    """Validate a Back-Channel Logout Token (async) (Back-Channel Logout 1.0 §2.4).
+
+    Runs the standard async JWT validation via ``validate_token`` (signature,
+    ``iss``, ``aud``, ``iat`` and ``exp`` when present), then enforces the
+    Logout-Token-specific rules: required ``events`` member, ``sub``/``sid``
+    presence, and the ``nonce`` prohibition.
+
+    Args:
+        logout_token: The compact-serialized Logout Token JWT received at the
+            RP's ``backchannel_logout_uri``.
+        token_validation_config: Validation configuration. ``audience`` should
+            be the RP's ``client_id`` and ``issuer`` the OP issuer.
+        disco_doc_address: Discovery document address (required when
+            ``perform_disco`` is True).
+        http_client: Optional managed async HTTP client (see ``validate_token``
+            for the cache-bypass caveats).
+
+    Returns:
+        dict: The decoded and validated Logout Token claims.
+
+    Raises:
+        TokenValidationException: If standard JWT validation fails (bad
+            signature, wrong issuer/audience, expired token).
+        LogoutTokenValidationException: If a Logout-Token-specific rule fails.
+        ConfigurationException: If the configuration is invalid.
+    """
+    claims = await validate_token(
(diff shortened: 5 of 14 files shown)


jamescrowley321:
## Review artifact: review-acceptance.md

## Review: Acceptance Auditor

Scope: Issue #442 (Back-Channel Logout RP — `validate_logout_token`), planning epic
`epic-0e-spec-logout.md` story S.10b. The acceptance criteria are the conformance
IDs LOGOUT-004, LOGOUT-005, LOGOUT-006, LOGOUT-007, LOGOUT-009, LOGOUT-010.
LOGOUT-008 (front-channel) is dropped/out-of-scope and is not evaluated.

Per the caller's guidance: `validate_logout_token` /
`validate_logout_token_claims` is a pure JWT-validation function of THIS library
(not a cross-service endpoint of this repo), so unit-tested pure-logic + sync/async
wrapper coverage is the accepted primary gate. The conformance HTTP receiver is a
thin test aid and live/hosted conformance runs are deferred (noted in PR).

### PASS
- [LOGOUT-004] Parse and validate a well-formed Logout Token (`iss`, `sub`, `aud`, `iat`, `jti`, `events` with the backchannel-logout member) — accepted.
  - Implemented: `src/py_identity_model/core/logout_logic.py:309` (`validate_logout_token_claims`), wrapped at `src/py_identity_model/sync/logout.py:490` and `src/py_identity_model/aio/logout.py:230`. Standard JWT checks delegated to `validate_token` (`src/py_identity_model/sync/token_validation.py:475`, returns claim dict).
  - Tested: `src/tests/unit/test_logout_token_validation.py:870` (pure) and `:927` (full signature+discovery flow via respx); async parity `src/tests/unit/test_aio_logout_token_validation.py:629`. (tier: unit — pure claim logic + signed-JWT flow with mocked JWKS/discovery)

- [LOGOUT-005] Reject Logout Token missing the `events` claim.
  - Implemented: `core/logout_logic.py:331` (`_REQUIRED_CLAIMS` loop, includes `events`).
  - Tested: `test_logout_token_validation.py:874` (pure) and `:984` (full flow); async `test_aio_logout_token_validation.py:677`. (tier: unit)

- [LOGOUT-006] Reject Logout Token whose `events` claim lacks the `http://schemas.openid.net/event/backchannel-logout` member.
  - Implemented: `core/logout_logic.py:354-365` (rejects non-dict `events` and missing member).
  - Tested: `test_logout_token_validation.py:881` (wrong member) and `:889` (events not an object). (tier: unit)

- [LOGOUT-007] Reject expired Logout Token (`exp` in the past).
  - Implemented: `exp` enforced by the delegated `validate_token` path (`decode_with_config`/`validate_claims`); wrapper preserves the raised exception.
  - Tested: `test_logout_token_validation.py:941` (full signed flow, expects `TokenExpiredException`); positive `:955` confirms `exp` is optional; async parity `test_aio_logout_token_validation.py:643`. (tier: unit — exercises real JWT decode + expiry check)

- [LOGOUT-009] Reject Logout Token with neither `sub` nor `sid`.
  - Implemented: `core/logout_logic.py:346-350`.
  - Tested: `test_logout_token_validation.py:895` (neither → reject) and `:902` (`sid`-only positive case). (tier: unit)

- [LOGOUT-010] Reject Logout Token containing a `nonce` claim.
  - Implemented: `core/logout_logic.py:339-343`.
  - Tested: `test_logout_token_validation.py:909` (pure) and `:968` (nonce rejected AFTER a valid signature, proving the wrapper enforces it); async `test_aio_logout_token_validation.py:659`. (tier: unit)

### FAIL
- (none)

### PARTIAL
- (none)

### Additional issue-scope items (beyond the six numbered ACs)
These are in the #442 scope statement; recorded for completeness, none blocking.

- Discovery fields `backchannel_logout_supported` / `backchannel_logout_session_supported` — PASS. Implemented `core/models.py:390-392` + `core/response_processors.py:405-411`; parsed set updated `core/models.py:381-382`. Tested `src/tests/unit/test_discovery_logout_fields.py` (populated / absent-defaults-None / `false` preserved). (tier: unit)
- `MUST contain iss, aud, iat, jti, events` — PASS. `jti` requirement covered by `test_logout_token_validation.py:915`; `iss`/`aud`/`iat` covered both by `_REQUIRED_CLAIMS` and standard JWT validation.
- `MUST reject nonce` / `sub`-or-`sid` — covered by LOGOUT-010 / LOGOUT-009 above.
- Sync + async parity — PASS. Both wrappers share the single pure `validate_logout_token_claims`; async suite mirrors the sync assertions.
- Public API export — PASS. `validate_logout_token` and `LogoutTokenValidationException` exported from `__init__.py`, `sync/__init__.py`, `aio/__init__.py`.

- **`SHOULD honor typ: logout+jwt`** — NOT implemented (observation, non-blocking). The issue lists this as a SHOULD, and there is no numbered LOGOUT-* AC for it, so it does not fail an acceptance criterion. Flagged so it is a conscious omission: no check exists that the JWT header `typ` is `logout+jwt`, and no test asserts one. Recommend a follow-up if strict §2.4 step 6 conformance is later desired.
- **Integration test / example** — the issue text mentions "integration test; example." No real-provider integration test is added; the conformance harness receiver (`conformance/app.py` `/backchannel-logout` + `_receive_backchannel_logout`, config `conformance/configs/backchannel-logout-rp.json`) serves as the example/harness surface, with live conformance runs deferred per the caller's note. Accepted, non-blocking.

### SCOPE CREEP
- (none) — the conformance receiver, `_last_flow_context` capture in `/authorize`, and the plan config file are all in-scope per the issue's "Conformance tests targeted" and the caller's note that the HTTP receiver is part of scope.

### Architecture Violations
- (none) — business logic is pure and lives in `core/logout_logic.py`; sync/async layers are thin wrappers over `validate_token`, matching the mandated core/sync/aio separation. Constants sourced from `oidc_constants.py` (`Events.BACK_CHANNEL_LOGOUT`, `BackChannelLogoutRequest.LOGOUT_TOKEN`). New exception inherits `TokenValidationException` → `PyIdentityModelException` per the exception hierarchy convention.

### Summary
- Total ACs: 6
- Pass: 6 | Fail: 0 | Partial: 0

jamescrowley321:
## Review artifact: review-blind.md

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`conformance/app.py:38-43`] `/authorize` overwrites the process-global `_last_flow_context` on every call with no locking or per-flow keying. Two overlapping certification flows clobber each other, so a Back-Channel Logout Token can be validated against the wrong `issuer`/`client_id`/`disco_address` — silently producing a false reject (or accepting a token bound to a different OP), yielding wrong/flaky test results.
- [`conformance/app.py:79-83`] TOCTOU on `_last_flow_context`: after the `is None` guard at line 67 the code re-reads the global three separate times (`["client_id"]`, `["issuer"]`, `["disco_address"]`). A concurrent `/authorize` write between those reads can splice fields from two different flow contexts into a single validation config.
- [`conformance/configs/backchannel-logout-rp.json:10`] A client secret (`"conformance-rp-secret"`) is committed in plaintext to the repo. Even for a test fixture this is a credential checked into source control; it should come from an env var or a gitignored file.

### NITPICK
- [`src/py_identity_model/core/logout_logic.py:346`] The sub/sid rule uses `claims.get("sub") is None and claims.get("sid") is None`, so a present-but-empty `sub`/`sid` (e.g. `""`) passes the "must identify the subject" check even though it carries no identifying value.
- [`conformance/app.py:113`] The `/backchannel-logout` handler has no `jti`/replay dedup, so the same Logout Token can be replayed and re-accepted; acceptable for the harness but worth noting against §2.4/§2.6 recommendations.

jamescrowley321:
## Review artifact: review-blind-delta.md

Scope: fix diff at `/tmp/pim-BCL/.claude/review-fix-diff.patch` only.

### MUST FIX
No findings.

### SHOULD FIX
No findings.

The three previously-raised items are all correctly resolved:

1. **TOCTOU snapshot (`app.py:866`)** — FIXED CORRECTLY. `flow_context = _last_flow_context`
   is bound once; the None-check (867) and the three field reads (`client_id` 879,
   `issuer` 880, `disco_address` 883) all read the local. Because `/authorize`
   *reassigns* `_last_flow_context` to a fresh dict (never mutates in place), the
   snapshotted reference is a single self-consistent flow — the field-splice hazard
   is gone. No `global` declaration needed since the receiver only reads.

2. **Unlocked global write in `/authorize`** — rationale holds for the changed code.
   With the snapshot in place, the only residual race is a whole-dict swap landing
   between the snapshot and use, which can substitute one *complete* flow context for
   another but can never splice fields. Acceptable under serial OIDF driving. No
   regression introduced by the fix.

3. **`except PyJWTError` (`app.py:903`)** — FIXED CORRECTLY. `jwt.exceptions.PyJWTError`
   is an independent exception hierarchy from `PyIdentityModelException`, so ordering
   it last neither masks the library `except` clauses above it (885/891/897) nor is
   shadowed by them. Returns 400 per Back-Channel Logout §2.5. Import path correct.
   No over-catch of unrelated errors: PyJWTError only covers JWT-shaped failures.

### NITPICK
- [`app.py:903-912`] `PyJWTError` is a broad base class (covers `PyJWKClientError`,
  `InvalidKeyError`, etc.). If the library ever surfaced a key/JWKS-fetch failure as a
  raw PyJWT error rather than wrapping it in a `PyIdentityModelException`, it would be
  reported to the OP as `"malformed token"` even though the token was fine. Harmless
  here (harness uses py-identity-model's own JWKS layer, which wraps its errors, and
  every branch returns 400 anyway), but the label could mislead when reading logs.
- [`app.py:906`] Comment says the malformed token "trips PyJWT's header decode inside
  ``validate_token``" — this path is `validate_logout_token`, not `validate_token`.
  Presumably it delegates internally, but the reference is slightly imprecise.

jamescrowley321:
## Review artifact: review-edge-case.md

## Review: Edge Case Hunter

### Findings

| `conformance/app.py:869` | OP posts a non-empty but malformed `logout_token` (e.g. `"garbage"`, not a JWT) | `except (PyIdentityModelException, PyJWTError) as exc:` (import `from jwt.exceptions import PyJWTError`) | [WRONG] `validate_token`→`extract_jwt_header_fields`→PyJWT `get_unverified_header` raises `jwt.exceptions.DecodeError` (not a `PyIdentityModelException`), uncaught → HTTP 500 instead of spec-mandated 400 |
| `conformance/app.py:873` | Concurrent `/authorize` replaces `_last_flow_context` between the None-check and the three key reads in a threadpool worker | `ctx = _last_flow_context` (snapshot once, then read ctx[...]) | [WRONG] logout token validated against a different flow's issuer/client_id/disco than intended (mixed context) |

### Summary
- Unhandled paths found: 2

jamescrowley321:
## Review artifact: review-edge-case-delta.md

## Review: Edge Case Hunter (delta)

Scope: fix diff at `.claude/review-fix-diff.patch` against `conformance/app.py:_receive_backchannel_logout`.

### Verification of previously-addressed findings

**Finding 1 — [WRONG] malformed `logout_token` → 500 instead of 400: FIXED, correct & complete.**
- Traced the live path: `validate_logout_token` (`src/py_identity_model/sync/logout.py:55`) → `validate_token` (`sync/token_validation.py:518`). With `perform_disco=True`, `_discover_and_resolve_key` runs discovery + JWKS first, then calls `extract_jwt_header_fields` (`token_validation.py:315`/`325`) → `get_unverified_header` (`core/parsers.py:92`), which raises PyJWT `DecodeError`.
- `DecodeError` is a `PyJWTError` subclass from the `jwt.exceptions` hierarchy and is NOT wrapped in a `PyIdentityModelException` at that call site — confirmed no intervening try/except swallows/wraps it in `validate_token`. So the new `except PyJWTError` (app.py:903) is the correct and only catch, returning 400.
- Handler ordering is sound: `LogoutTokenValidationException` → `TokenValidationException` → `PyIdentityModelException` (specific→base within the py_identity_model tree), then `PyJWTError` (disjoint hierarchy, so no shadowing regardless of position). Import at app.py:22 is valid (PyJWT is a hard dependency).
- Reachability confirmed: empty/`None` token is short-circuited earlier (app.py:855) and non-str is coerced to None (app.py:931), so only a non-empty, non-well-formed string reaches the header decode — exactly the trigger.

**Finding 2 — [WRONG] TOCTOU splice of `_last_flow_context`: FIXED, correct & complete.**
- `flow_context = _last_flow_context` snapshots the reference once (app.py:866); the None-check and all three field reads (`client_id`, `issuer`, `disco_address` at app.py:879-883) now use the local.
- Confirmed `/authorize` rebinds the global to a brand-new dict literal (app.py:479-484) rather than mutating in place, so a concurrent rebind cannot alter the already-bound local snapshot. The splice window is fully closed.

### Findings (new regressions in the fix diff)

| — | — | — | — |

### Summary
- Unhandled paths found: 0

No unhandled edge cases found. Both previously-addressed findings are correctly and completely fixed; the fix diff introduces no new unhandled path.

jamescrowley321:
## Review artifact: review-sentinel.md

None.

### WARN (should fix)
None.

### INFO (acceptable risk)

- `src/py_identity_model/core/logout_logic.py:1-72` — Logout-Token rule ordering is correct: `validate_logout_token_claims` runs strictly *after* `validate_token` (see `sync/logout.py:522-528` and `aio/logout.py:262-268`), so signature, `iss`, `aud`, `iat` and `exp` are cryptographically verified before any logout-specific claim is trusted. There is no path that reaches the claim rules on an unverified token — the ID-token-confusion vector (replaying an ID token as a logout token) is blocked by the mandatory `events` member check plus the `nonce`-prohibition check. Rules cannot be individually bypassed: all five checks are unconditional and each raises before returning.

- `src/py_identity_model/core/logout_logic.py:346` — The `sub`/`sid` presence rule uses `claims.get("sub") is None and claims.get("sid") is None`. An empty-string `sub`/`sid` (`""`) would satisfy the rule. Not exploitable: the value still had to survive signature + issuer validation from a trusted OP, and an OP emitting an empty subject is a provider bug, not an attack primitive. Spec §2.4 only requires presence, which is met.

- `src/py_identity_model/sync/logout.py:490` / `aio/logout.py:230` — The wrapper does not force `token_validation_config.audience` to be set. If a caller omits `audience`, the underlying `validate_token` skips `aud` verification and a logout token minted for a *different* client of the same OP would be accepted. This is the standard `validate_token` contract (identical to ID-token validation) and the docstrings direct callers to set `audience` to the RP `client_id`; it is a caller-responsibility / defense-in-depth item, not a bypass introduced by this change. The conformance harness sets it correctly (`app.py:79`).

- Algorithm handling (reviewed via `core/parsers.py:find_key_by_kid` + `core/jwt_helpers.py`) — no `alg=none` / algorithm-confusion vector. The decode `algorithms` list is derived from the resolved JWKS key (the key's declared `alg`, else `RS256`), never from the attacker-controlled JWT header, and `_validate_key_alg_consistency` rejects key-type/alg mismatches. A forged `alg:none` token is rejected because `none` is never in the allowed-algorithms list.

- `conformance/app.py:26,38-43,67-72` — The `/backchannel-logout` receiver recovers `issuer`/`client_id`/`disco_address` from a process-global `_last_flow_context` set by the most recent `/authorize`. This is mutable global state with no per-session keying and could cross-contaminate between concurrent/interleaved flows, but this is a single-tenant certification harness driven serially by the conformance suite, and the values are supplied by the suite's own authorize call (not attacker-controlled). Acceptable for a test harness. The endpoint is intentionally unauthenticated at the HTTP layer (auth is the signed logout token) and `require_https=False` is required for the self-signed conformance environment — both correct for this context.

- `conformance/app.py:104-108` — On success the receiver logs `sub`/`sid` claim values. These are session/subject identifiers (PII-adjacent) but confined to the test harness; no token bytes or secrets are logged, and error paths log only exception messages, not the raw `logout_token`.

### Summary
- BLOCK: 0 | WARN: 0 | INFO: 6
- Overall: PASS

jamescrowley321:
## Review artifact: review-viper.md

## Review: Red Team (Viper)

### Attack Surface
- `core/logout_logic.py:validate_logout_token_claims` — new pure claim gate (events member, sub/sid, nonce prohibition, required claims). Runs *after* standard JWT validation.
- `sync/logout.py` / `aio/logout.py:validate_logout_token` — thin wrappers that call `validate_token` then the claim gate. The security of the whole feature rests on `validate_token` actually enforcing signature/iss/aud.
- `conformance/app.py:/backchannel-logout` + `_receive_backchannel_logout` — new unauthenticated POST receiver that validates an OP-supplied `logout_token` using issuer/client_id recovered from module-global `_last_flow_context`.
- `core/models.py` / `core/response_processors.py` — new discovery flags (`backchannel_logout_supported`, `backchannel_logout_session_supported`), parse-only, no trust decision. Not a security surface.

### Verification of the reused validation path (as requested)
Traced `validate_token` (sync `sync/token_validation.py:475`, async `aio/token_validation.py:498`):
- **Signature IS enforced.** `decode_with_config` → `decode_and_validate_jwt` → PyJWT `decode()` with a real JWK. `_sanitize_options` (`core/jwt_helpers.py:39`) hard-blocks any attempt to set `verify_signature`/`verify_exp`/`verify_nbf`/`verify_iat` to False. A forged/unsigned token raises `SignatureVerificationException`. **Forged/unsigned logout tokens are rejected.**
- **`alg=none` / alg-confusion blocked.** The allowed algorithm list handed to PyJWT is `[alg]` where `alg` is derived from the resolved JWK (`find_key_by_kid`), never from the attacker's header. `none`/`HS256`-with-RSA-pubkey attempts fail because the header alg is not in the allowed list, and `_validate_key_alg_consistency` (`core/parsers.py:34`) rejects kty/alg mismatches.
- **`iss` bound to discovery.** `decode_with_config(..., disco_doc_response.issuer)` forces the issuer actually checked to be the discovery document's issuer, not attacker-influenced.
- **`aud` enforced when configured.** Config passes `audience=client_id`; PyJWT verifies it. (Caveat in LOW-2 below.)
- **Token-type confusion (ID/access token → logout token) is blocked** by two independent gates: the required `events` member (ID/access tokens don't carry it → rejected, LOGOUT-005/006) and the `nonce` prohibition (an implicit-flow ID token carrying `nonce` → rejected, LOGOUT-010). Sync/async parity is real because both wrappers call the same pure `validate_logout_token_claims`.

The core token-type-confusion / forgery / bypass questions in the brief all resolve to **not exploitable**.

### Findings

#### LOW — Back-channel logout replay: `exp` optional and no `jti` de-dup enables forced-logout replay
- **Location**: `core/logout_logic.py:309` (`_REQUIRED_CLAIMS` omits `exp`; no `jti` uniqueness tracking); confirmed by test `test_logout_token_without_exp_accepted`.
- **Attack scenario**:
  1. Attacker observes/captures one legitimate, correctly-signed `logout_token` for victim (e.g. from a proxy log, or a malicious downstream that received a broadcast logout).
  2. Because the library does not require `exp` and implements no `jti` replay cache (spec §2.4 step 6 is optional but recommended), a captured token with no `exp` is valid indefinitely.
  3. Attacker re-POSTs it to the RP's `backchannel_logout_uri` repeatedly to force the victim's session termination at will.
- **Prerequisites**: Ability to capture one valid signed logout token; network reach to the RP receiver. Impact is limited to availability (forced logout), not confidentiality/integrity.
- **Remediation**: Document that callers MUST implement `jti` replay protection, and consider surfacing a helper. Optionally treat a missing `exp` as a caller-configurable rejection. At minimum add a doc note that `validate_logout_token` performs no replay protection.

#### LOW — `aud` value is required-present but not required-to-match when caller omits `audience`
- **Location**: `core/logout_logic.py:306` lists `"aud"` in `_REQUIRED_CLAIMS` (presence only); actual value check is delegated to `validate_token`, which does NOT verify `aud` when `TokenValidationConfig.audience is None`.
- **Attack scenario**:
  1. An RP integrates `validate_logout_token` but leaves `audience=None` in the config (the wrapper does not force it; docstring only says it "should" be the client_id).
  2. Attacker replays a validly-signed logout token issued by the *same OP* for a *different* RP (different `aud`).
  3. Signature + `iss` pass; `aud` present-but-unchecked → the token is accepted, logging out a session it was not issued for (cross-RP forced logout at a shared OP).
- **Prerequisites**: Victim RP misconfigured with `audience=None`; attacker holds a valid logout token for another RP at the same issuer.
- **CVSS v3.1**: 3.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L)
- **Remediation**: Since §2.4 mandates `aud` validation "same as ID Tokens", have `validate_logout_token`/the claim gate reject when the effective config would not verify `aud` (i.e. require `token_validation_config.audience` to be set), rather than only asserting `aud` is present.

#### LOW — Conformance receiver disables TLS verification and `require_https=False` (MITM logout forgery in the harness only)
- **Location**: `conformance/app.py:_receive_backchannel_logout` (`require_https=False`) using the shared `_http_client` that skips TLS verification, plus discovery/JWKS fetched over the recovered `disco_address`.
- **Attack scenario**:
  1. Attacker in a network MITM position between the harness and the OP.
  2. Because discovery/JWKS are fetched with TLS verification off and `require_https=False`, attacker serves a forged discovery doc + attacker-controlled JWKS.
  3. Attacker then forges a `logout_token` signed with their own key; it validates against the spoofed JWKS and the receiver returns 200.
- **Prerequisites**: MITM on the harness↔OP path. This is test-only infrastructure (the suite uses self-signed certs by design), so the exposure does not reach production code. Not exploitable in the shipped library.
- **CVSS v3.1**: 3.0 (AV:A/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N) — harness scope only.
- **Remediation**: None required for the library. If desired, isolate the TLS-skipping client to the conformance network and document that the receiver must never be reused outside the harness.

#### INFO — `events` member value type not checked (spec-conformant, not a vuln)
- **Location**: `core/logout_logic.py:360` checks the member key is present but not that `events[BACKCHANNEL_LOGOUT_EVENT]` is a JSON object.
- Spec §2.4 only requires the value SHOULD be an empty object and MAY carry other values, so accepting any value is compliant. No attack path; noted for completeness.

#### INFO — `_last_flow_context` is an unlocked module global mutated in async `/authorize` and read in `/backchannel-logout`
- **Location**: `conformance/app.py:26` / `:472`. Concurrent flows can interleave, but the receiver still enforces signature against whichever issuer/JWKS it reads, so the worst case is a validation failure (400), not acceptance of a bad token. Correctness nit in test tooling, not a security boundary.

### Summary
- Attack surface elements: 4 (logout claim gate, sync/async wrappers, conformance receiver, discovery flags)
- Findings: 0 critical, 0 high, 0 medium, 3 low (+2 informational)
- Overall: PASS — the core model (signature/iss/aud reuse, alg-confusion defense, events/sub-sid/nonce rules, token-type-confusion prevention) is correctly enforced and holds up under attack. The LOW items are hardening/documentation gaps and a test-only MITM caveat; none is remotely exploitable against the shipped library with a correctly-configured caller.

Issue #214, opened 2026-03-24
Implement RP-Initiated Logout (End Session)


jamescrowley321:
## Goal

Implement OpenID Connect RP-Initiated Logout for ending user sessions at the identity provider.

## Context

After token revocation (#17), applications also need to end the user's session at the IdP. This is the standard OIDC logout mechanism — the RP redirects the user to the IdP's `end_session_endpoint` with an `id_token_hint`.

## Specification References

- **OpenID Connect RP-Initiated Logout 1.0** — [Spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
  - §2 RP-Initiated Logout — Redirect to `end_session_endpoint`
  - §2.1 Logout Request Parameters: `id_token_hint`, `client_id`, `post_logout_redirect_uri`, `state`
  - §3 Validation — IdP validates the id_token_hint and redirect URI
- **OpenID Connect Front-Channel Logout 1.0** — [Spec](https://openid.net/specs/openid-connect-frontchannel-1_0.html) — Front-channel logout via iframes
- **OpenID Connect Back-Channel Logout 1.0** — [Spec](https://openid.net/specs/openid-connect-backchannel-1_0.html) — Server-to-server logout notification via `logout_token`
- **OpenID Connect Discovery 1.0** — §3 Provider Metadata:
  - `end_session_endpoint`
  - `frontchannel_logout_supported`, `frontchannel_logout_session_supported`
  - `backchannel_logout_supported`, `backchannel_logout_session_supported`
- **OpenID Connect Session Management 1.0** — [Spec](https://openid.net/specs/openid-connect-session-1_0.html) — `session_state` parameter, check_session_iframe

## Duende IdentityModel Reference

- Duende OidcClient handles end session:
  - [`LogoutRequest`](https://github.com/DuendeSoftware/foss/tree/main/identity-model-oidc-client/src/IdentityModel.OidcClient) — End session request with `IdTokenHint`
  - `EndSessionUrl` construction from discovery document `end_session_endpoint`
- [`OidcConstants.EndSessionRequest`](https://github.com/DuendeSoftware/foss/blob/main/identity-model/src/IdentityModel/OidcConstants.cs) — Parameter constants
- [`RequestUrl`](https://github.com/DuendeSoftware/foss/blob/main/identity-model/src/IdentityModel/Client/RequestUrl.cs) — URL builder used for end session URL construction

## Tasks

[...]

PR #451, opened 2026-07-21
feat(logout): RP-initiated logout — build_end_session_url + state round-trip

jamescrowley321:
## Summary
Implements the OIDC **RP-Initiated Logout 1.0** relying-party profile (issue #214) for the next certification round.

- **`core/logout_logic.build_end_session_url(...)`** — pure builder for the OP end-session URL. Includes only non-`None` params (`id_token_hint`, `client_id`, `post_logout_redirect_uri`, `state`, plus `**extra_params`), rejects an empty endpoint and reserved-param collisions, strips any `#fragment`, and picks the `?`/`&` separator correctly (incl. a bare trailing `?`/`&`). Mirrors `build_authorization_url` — pure, no I/O, so it's imported directly into both `sync` and `aio` surfaces (no wrapper files, parity trivially satisfied).
- **`core/logout_logic.validate_post_logout_state(expected, returned)`** — constant-time (`hmac.compare_digest`) state check for the post-logout redirect; raises `LogoutStateValidationException` on missing-expected, missing-returned, or mismatch. Returns `None` on match.
- **`LogoutStateValidationException(ValidationException)`** — new exception (state is a CSRF concern, not a JWT/token error).
- **Discovery model** gains `end_session_endpoint` (`DiscoveryDocumentResponse` field + `_guarded_fields`; `build_discovery_response` mapping + `_endpoint_names` so it gets HTTPS/authority policy validation).
- Public exports from `py_identity_model`, `.sync`, `.aio`.
- **Conformance harness** (`conformance/app.py`): `/logout` builds the end-session URL and 302-redirects; `/post-logout-callback` validates `state`; `AuthSession` captures `id_token`/`end_session_endpoint`. New config `conformance/configs/rpinitiated-logout-rp.json` (plan `oidcc-client-rpinitiated-logout-certification-test-plan` — exact name to be confirmed at the hosted run).

Refs #214

Base: `feat/backchannel-logout` (BCL PR #450 still open — stacked, not `main`).

## Spec IDs covered
- **LOGOUT-001** — end-session URL with all params (`id_token_hint`/`post_logout_redirect_uri`/`state`/`client_id`).
- **LOGOUT-002** — end-session URL with only `id_token_hint`.
- **LOGOUT-003** — validate `state` on the post-logout redirect; raise on mismatch/missing.
- (LOGOUT-008 front-channel = **dropped**, not implemented.)

23 unit tests (`test_end_session_url.py`, `test_post_logout_state.py`, extended `test_discovery_logout_fields.py`).

## Review Findings Addressed
Reviewers: Blind + Edge + Acceptance + Sentinel + Viper @ `4389644`.
- **P0: 0.** Acceptance 4/4 AC PASS. Sentinel PASS (0 BLOCK/0 WARN/3 INFO harness-only). Viper PASS (0 crit/0 high/1 MEDIUM/2 LOW).
- **P1 fixed** (commit): (1) Edge `[WRONG]` — doubled `?`/`&` separator corrupting the first param key when the endpoint ends in a bare `?`/`&`; guarded + regression test. (2) Viper MEDIUM / Sentinel INFO — harness `/logout` logged the full end-session URL (embedding `id_token_hint` = full ID Token with PII); now logs only the bare end-session endpoint.
- **P1 skipped w/ rationale:** `_expected_logout_state`/`_last_logout_context` are process-global, not session-bound, never cleared — harness is serially cert-driven (single flow at a time), consistent with the BCL `_last_flow_context` precedent already cleared as harness-acceptable; reviewers scope these to "before lifting into production middleware."
- **P2 skipped:** empty-string-param emission (mirrors `build_authorization_url` `is not None` precedent), dup-param edge, trailing-bare-`&` fail-safe.
- Delta re-review (Edge + Viper, `4389644...00e8b38`): both PASS, no regressions.

## Test plan
- [x] Unit tests pass (`make test-unit`: 1268 passed, 95.61% cov)
- [x] Integration tests pass (`[skip-integration-tests: pure library fns + discovery field, fully covered by respx-mocked unit tests; no conftest/fixture/shared-util edits, no integration code path exercised]`)
- [x] Lint passes (`make lint` — ruff/format/pyrefly/pytest-80% all pass)
- [ ] CI passes

### DIFF
--- a/conformance/app.py
+++ b/conformance/app.py
@@ -995,7 +995,12 @@ def logout() -> RedirectResponse | JSONResponse:
         state=_expected_logout_state,
     )
 
-    logger.info("Redirecting to OP end-session endpoint: %s", end_session_url)
+    # Log only the endpoint, never the full URL: it carries id_token_hint (a
+    # complete ID Token with PII/subject claims) which must not land in logs.
+    logger.info(
+        "Redirecting to OP end-session endpoint: %s",
+        logout_context["end_session_endpoint"],
+    )
     return RedirectResponse(url=end_session_url, status_code=302)
 
 
--- a/conformance/app.py
+++ b/conformance/app.py
@@ -30,6 +30,7 @@ from py_identity_model import (
     TokenValidationConfig,
     UserInfoRequest,
     build_authorization_url,
+    build_end_session_url,
     clear_jwks_cache,
     generate_pkce_pair,
     get_discovery_document,
@@ -39,10 +40,12 @@ from py_identity_model import (
     request_authorization_code_token,
     validate_authorize_callback_state,
     validate_logout_token,
+    validate_post_logout_state,
     validate_token,
 )
 from py_identity_model.exceptions import (
     ConfigurationException,
+    LogoutStateValidationException,
     LogoutTokenValidationException,
     PyIdentityModelException,
     TokenValidationException,
@@ -198,6 +201,10 @@ class AuthSession:
     client_secret: str = ""
     redirect_uri: str = ""
     skip_userinfo: bool = False
+    # RP-Initiated Logout: the id_token issued to this session (used as
+    # id_token_hint) and the OP's end-session endpoint, captured in /callback.
+    id_token: str = ""
+    end_session_endpoint: str = ""
     # Per-test RP-log routing (recovered in /callback via the session)
     test_name: str = ""
     profile: str = ""
@@ -219,6 +226,17 @@ _http_client: HTTPClient | None = None
 # carries no flow parameters, so the receiver recovers the issuer (for signature
 # + iss validation) and client_id (audience) from the last flow it drove.
 _last_flow_context: dict[str, str] | None = None
+
+# Most recent completed-login context for RP-Initiated Logout: the OP's
+# end-session endpoint, the issued id_token (used as ``id_token_hint``), and the
+# client_id. Captured in /callback on a successful login; consumed by /logout,
+# which carries no flow parameters of its own.
+_last_logout_context: dict[str, str] | None = None
+
+# The ``state`` the harness sent to the end-session endpoint on the most recent
+# /logout redirect. Checked against the value the OP returns to
+# /post-logout-callback (RP-Initiated Logout 1.0 §2 state round-trip).
+_expected_logout_state: str | None = None
 
 
 def _get_http_client() -> HTTPClient:
@@ -785,6 +803,18 @@ def _handle_callback(request_url: str) -> HTMLResponse | JSONResponse:
             logger.warning("UserInfo exception: %s", exc)
             session.result["userinfo_error"] = str(exc)
 
+    # Record the id_token + end-session endpoint so a later /logout can build the
+    # RP-Initiated Logout URL. /logout carries no flow parameters of its own, so
+    # (like the back-channel receiver) it recovers them from the last flow driven.
+    session.id_token = id_token_jwt or ""
+    session.end_session_endpoint = disco.end_session_endpoint or ""
+    global _last_logout_context
+    _last_logout_context = {
+        "end_session_endpoint": session.end_session_endpoint,
+        "id_token": session.id_token,
+        "client_id": session.client_id,
+    }
+
     # Store successful result
     session.result["status"] = "success"
     session.result["id_token_claims"] = claims
@@ -931,6 +961,69 @@ async def backchannel_logout(request: Request) -> JSONResponse:
     if not isinstance(logout_token, str):
         logout_token = None
     return await run_in_threadpool(_receive_backchannel_logout, logout_token)
+
+
+@app.get("/logout", response_model=None)
+def logout() -> RedirectResponse | JSONResponse:
+    """Initiate RP-Initiated Logout (OpenID Connect RP-Initiated Logout 1.0 §2).
+
+    Builds the OP's end-session URL with the ``id_token_hint``, ``client_id``,
+    ``post_logout_redirect_uri`` and a fresh ``state`` via the library's
+    ``build_end_session_url``, remembers the ``state`` for the round-trip check,
+    and redirects the user agent to the OP. Returns 400 if no login flow has
+    completed yet (no end-session endpoint / id_token to log out with).
+    """
+    logout_context = _last_logout_context
+    if logout_context is None or not logout_context.get("end_session_endpoint"):
+        logger.error("REJECTED: /logout called before a successful login flow")
+        return JSONResponse(
+            status_code=400,
+            content={
+                "error": "no_logout_context",
+                "detail": "no completed login flow with an end_session_endpoint",
+            },
+        )
+
+    global _expected_logout_state
+    _expected_logout_state = secrets.token_urlsafe(32)
+
+    end_session_url = build_end_session_url(
+        logout_context["end_session_endpoint"],
+        id_token_hint=logout_context["id_token"] or None,
+        client_id=logout_context["client_id"] or None,
+        post_logout_redirect_uri=f"{RP_BASE_URL}/post-logout-callback",
+        state=_expected_logout_state,
+    )
+
+    logger.info("Redirecting to OP end-session endpoint: %s", end_session_url)
+    return RedirectResponse(url=end_session_url, status_code=302)
+
+
+@app.get("/post-logout-callback", response_model=None)
+def post_logout_callback(state: str | None = Query(default=None)) -> HTMLResponse:
+    """Post-logout redirect target (RP-Initiated Logout 1.0 §2 state round-trip).
+
+    Validates the ``state`` the OP returns against the value sent to the
+    end-session endpoint using the library's ``validate_post_logout_state``.
+    Responds 200 on a matching round-trip and 400 on a missing/mismatched state.
+    """
+    try:
+        validate_post_logout_state(_expected_logout_state, state)
+    except LogoutStateValidationException as exc:
+        logger.error("REJECTED: post-logout state validation failed: %s", exc.message)
+        return HTMLResponse(
+            content=(
+                "<h1>Logout State Validation Failed</h1>"
+                f"<p>{html.escape(str(exc.message))}</p>"
+            ),
+            status_code=400,
+        )
+
+    logger.info("ACCEPTED: post-logout state validated by py-identity-model")
+    return HTMLResponse(
+        content="<h1>Logout Successful</h1><p>State validated.</p>",
+        status_code=200,
+    )
 
 
 @app.get("/results/{test_id}")
--- a/conformance/configs/rpinitiated-logout-rp.json
+++ b/conformance/configs/rpinitiated-logout-rp.json
@@ -1,0 +1,15 @@
+{
+  "plan_name": "oidcc-client-rpinitiated-logout-certification-test-plan",
+  "variant": {
+    "client_registration": "static_client",
+    "request_type": "plain_http_request"
+  },
+  "alias": "py-identity-model-rpinitiated-logout-rp",
+  "client": {
+    "client_id": "conformance-rp",
+    "client_secret": "conformance-rp-secret",
+    "post_logout_redirect_uris": [
+      "http://localhost:8888/post-logout-callback"
+    ]
+  }
+}
--- a/src/py_identity_model/__init__.py
+++ b/src/py_identity_model/__init__.py
@@ -15,6 +15,7 @@ from .exceptions import (
     InvalidAudienceException,
     InvalidIssuerException,
     JwksException,
+    LogoutStateValidationException,
     LogoutTokenValidationException,
     NetworkException,
     PyIdentityModelException,
@@ -75,6 +76,7 @@ from .sync import (
     UserInfoResponse,
     build_authorization_url,
     build_dpop_headers,
+    build_end_session_url,
     build_jar_authorization_url,
     clear_discovery_cache,
     clear_jwks_cache,
@@ -105,6 +107,7 @@ from .sync import (
     validate_fapi_client_config,
     validate_fapi_discovery,
     validate_logout_token,
+    validate_post_logout_state,
     validate_token,
 )
 
@@ -157,6 +160,7 @@ __all__ = [
     "JwksException",
     "JwksRequest",
     "JwksResponse",
+    "LogoutStateValidationException",
     "LogoutTokenValidationException",
     "NetworkException",
     # Client authentication (private_key_jwt)
@@ -191,6 +195,7 @@ __all__ = [
     "ValidationException",
     "build_authorization_url",
     "build_dpop_headers",
+    "build_end_session_url",
     "build_jar_authorization_url",
     "clear_discovery_cache",
     "clear_jwks_cache",
@@ -223,5 +228,6 @@ __all__ = [
     "validate_fapi_client_config",
     "validate_fapi_discovery",
     "validate_logout_token",
+    "validate_post_logout_state",
     "validate_token",
 ]
--- a/src/py_identity_model/aio/__init__.py
+++ b/src/py_identity_model/aio/__init__.py
@@ -49,6 +49,10 @@ from ..core.fapi import (
     validate_fapi_discovery,
 )
 from ..core.jar import build_jar_authorization_url, create_request_object
+from ..core.logout_logic import (
+    build_end_session_url,
+    validate_post_logout_state,
+)
 from ..core.models import BaseRequest, BaseResponse, PrivateKeyJwt
 from ..core.pkce import (
     generate_code_challenge,
@@ -183,6 +187,7 @@ __all__ = [
     "UserInfoResponse",
     "build_authorization_url",
     "build_dpop_headers",
+    "build_end_session_url",
     "build_jar_authorization_url",
     "compute_ath",
     "create_dpop_proof",
@@ -211,5 +216,6 @@ __all__ = [
     "validate_fapi_client_config",
     "validate_fapi_discovery",
     "validate_logout_token",
+    "validate_post_logout_state",
     "validate_token",
 ]
--- a/src/py_identity_model/core/logout_logic.py
+++ b/src/py_identity_model/core/logout_logic.py
@@ -192,8 +192,13 @@ def build_end_session_url(  # noqa: PLR0913  # RP-Initiated Logout 1.0 §2 defin
     if not params:
         return end_session_endpoint
 
+    query_string = urlencode(params)
+    # A bare trailing "?" or "&" already introduces the query component; appending
+    # another separator would corrupt the first parameter key (e.g. "logout??a=1").
+    if end_session_endpoint.endswith(("?", "&")):
+        return f"{end_session_endpoint}{query_string}"
     separator = "&" if parsed.query else "?"
-    return f"{end_session_endpoint}{separator}{urlencode(params)}"
+    return f"{end_session_endpoint}{separator}{query_string}"
 
 
 def validate_post_logout_state(
--- a/src/py_identity_model/core/logout_logic.py
+++ b/src/py_identity_model/core/logout_logic.py
@@ -1,15 +1,29 @@
-"""Back-Channel Logout token validation logic.
+"""Logout logic — Back-Channel and RP-Initiated Logout.
 
-Pure, protocol-agnostic claim validation shared by the synchronous and
-asynchronous logout wrappers. Contains no I/O: the standard JWT validation
-(signature, ``iss``, ``aud``, ``iat``, ``exp``) is performed by the existing
-``validate_token`` path; this module enforces only the additional rules that
-are unique to Logout Tokens (OpenID Connect Back-Channel Logout 1.0 §2.4).
+Pure, protocol-agnostic logic shared by the synchronous and asynchronous
+surfaces. Contains no I/O:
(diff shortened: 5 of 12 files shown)


jamescrowley321:
## Review artifact: acceptance

## Review: Acceptance Auditor

Scope (per authoritative logout spec, epic-0e-spec-logout.md, Story S.10a): RP-Initiated
Logout profile = **LOGOUT-001, LOGOUT-002, LOGOUT-003** plus adding `end_session_endpoint`
to the discovery model. LOGOUT-008 (front-channel) is explicitly DROPPED/out-of-scope and is
not evaluated. Back-channel test IDs (LOGOUT-004..007/009/010) are pre-existing and out of scope.

ACs enumerated from the spec:
- AC-1 = LOGOUT-001: Build RP-Initiated logout URL with all params (`id_token_hint`,
  `post_logout_redirect_uri`, `state`, `client_id`) appended as query values.
- AC-2 = LOGOUT-002: Build logout URL with only `id_token_hint` — no other params present.
- AC-3 = LOGOUT-003: Validate `state` round-trip on post-logout redirect; raise on mismatch.
- AC-4 = Discovery: `end_session_endpoint` parsed onto `DiscoveryDocumentResponse`.

### PASS
- [AC-1] LOGOUT-001 build end-session URL with all params — implemented in
  `build_end_session_url` at `src/py_identity_model/core/logout_logic.py:119` (params ordered
  and appended via `urlencode`), tested at `src/tests/unit/test_end_session_url.py:29`
  (`test_all_params` asserts scheme/netloc/path plus all four params) (tier: unit — the
  library surface is a pure string builder, so unit is the correct and sufficient tier).
- [AC-2] LOGOUT-002 build URL with only `id_token_hint` — same builder emits only supplied
  (non-`None`) params, tested at `src/tests/unit/test_end_session_url.py:52`
  (`test_id_token_hint_only` asserts `id_token_hint` present and `client_id`/
  `post_logout_redirect_uri`/`state` absent) (tier: unit).
- [AC-3] LOGOUT-003 validate `state` round-trip, raise on mismatch — implemented in
  `validate_post_logout_state` at `src/py_identity_model/core/logout_logic.py:199`
  (constant-time `hmac.compare_digest`, raises `LogoutStateValidationException` on
  missing/mismatch), tested at `src/tests/unit/test_post_logout_state.py:34`
  (`test_matching_state_returns_none`) and `:39` (`test_mismatched_state_raises`), plus
  missing/empty expected and returned cases at `:44`–`:61` (tier: unit — pure validation
  function; the negative-path tests would fail if the check were removed).
- [AC-4] Discovery `end_session_endpoint` — field added to `DiscoveryDocumentResponse` at
  `src/py_identity_model/core/models.py:537` and parsed at
  `src/py_identity_model/core/response_processors.py:277`
  (`build_discovery_response`), tested at
  `src/tests/unit/test_discovery_logout_fields.py:79` (`test_end_session_endpoint_populated`)
  and `:100` (`test_end_session_endpoint_absent_default_none`), driven through the real
  `get_discovery_document` parse path with a respx-mocked HTTP response (tier: unit with HTTP
  parse exercised via respx — the standard pattern for discovery parsing in this codebase).

### FAIL
- (none)

### PARTIAL
- (none)

### SCOPE CREEP
- `src/py_identity_model/core/logout_logic.py:124-125` — `build_end_session_url` adds
  `logout_hint` and `ui_locales` keyword params beyond the four AC-listed parameters. Both are
  legitimate RP-Initiated Logout 1.0 §2 parameters and are tested
  (`test_end_session_url.py:63` `test_logout_hint_and_ui_locales`); non-blocking, spec-aligned.
- `src/py_identity_model/core/logout_logic.py:127,161-165` — `**extra_params` passthrough with
  a reserved-parameter collision guard (`_RESERVED_END_SESSION_PARAMS`). Not required by any AC;
  tested (`test_extra_params_passthrough`). Non-blocking hardening.
- `conformance/app.py:78-138` — new `/logout` and `/post-logout-callback` FastAPI endpoints and
  `conformance/configs/rpinitiated-logout-rp.json`. This is the conformance-certification harness
  wiring for the RP-Initiated Logout plan, not one of the three library test-ID ACs. Supporting
  infrastructure for the profile rather than problematic scope creep; not independently
  unit-tested (harness code, exercised by the external conformance suite). Non-blocking.

### Architecture Violations
- (none) — Pure functions live in `core/logout_logic.py` and are imported directly into both the
  sync (`src/py_identity_model/sync/__init__.py`) and async
  (`src/py_identity_model/aio/__init__.py`) surfaces plus the top-level package, mirroring the
  established `build_authorization_url` / `validate_authorize_callback_state` pattern. Sync/async
  export parity is asserted by `test_public_export_identity` in both new test modules, satisfying
  the "both sync and async contexts" expectation. New `LogoutStateValidationException` correctly
  subclasses `ValidationException` (not `TokenValidationException`), verified at
  `test_post_logout_state.py:57`.

### Summary
- Total ACs: 4
- Note: the broader issue #214 also lists integration tests against a real IdP, a FastAPI
  example, and documentation. Per the caller's authoritative scope (LOGOUT-001/002/003 +
  discovery field), those are out of scope for this review and are not counted as failures.
- All 23 unit tests in the three affected test modules pass locally.

jamescrowley321:
## Review artifact: blind

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`conformance/app.py:99`] `_expected_logout_state` is a process-wide module global written in `/logout` and read in `/post-logout-callback`. Two overlapping logout flows overwrite each other's expected state, so the second flow's `/post-logout-callback` validates against the wrong value — producing false accept/reject. There is no per-session binding of the state.
- [`conformance/app.py:123`] `_expected_logout_state` is never cleared after a successful `validate_post_logout_state`. The captured CSRF secret stays valid indefinitely, so the same `state` can be replayed against `/post-logout-callback` to get a 200 "Logout Successful" repeatedly after the flow has already completed.
- [`conformance/app.py:64`] `_last_logout_context` is a module global overwritten on every successful `/callback`. When more than one login flow completes before `/logout` is hit, only the most recent id_token/end_session_endpoint survives; a `/logout` after concurrent flows logs out an arbitrary (last-wins) session rather than the intended one.

### NITPICK
- [`core/logout_logic.py:373`] Parameters are included when `is not None`, so an empty string (e.g. `id_token_hint=""` or `state=""`) is emitted as an empty query value (`state=`). This is inconsistent with `validate_post_logout_state`, which treats an empty `state` as invalid — the builder accepts what the validator rejects.
- [`core/logout_logic.py:390`] The separator is chosen from `parsed.query`, but if `end_session_endpoint` already contains a query parameter that collides with one being added (e.g. endpoint ends in `?state=x` and `state=` is also passed), the resulting URL carries duplicate parameters with no de-duplication.
- [`conformance/app.py:100`] `_expected_logout_state` is assigned before `build_end_session_url` is called; the reserved-endpoint check above guarantees no exception here, but the global is mutated prior to confirming the redirect URL was built, leaving stale state if the builder ever raised.

jamescrowley321:
## Review artifact: edge-case

## Review: Edge Case Hunter

### Findings

| `src/py_identity_model/core/logout_logic.py:390` | `end_session_endpoint` ends with a bare `?` (empty query) | `separator = "&" if end_session_endpoint.rstrip().endswith("?") or parsed.query else "?"` | [WRONG] Emits `logout??id_token_hint=...`; first param key corrupted to `?id_token_hint` |

### Summary
- Unhandled paths found: 1

jamescrowley321:
## Review artifact: edge-delta

## Review: Edge Case Hunter (DELTA re-review)

Scope: fix diff `/tmp/pim-RPL/.claude/review-delta.patch` only.

### Prior finding — RESOLVED

The previously-reported [WRONG] on `build_end_session_url` (a bare trailing `"?"`
producing `"logout??id_token_hint=..."`) is correctly fixed:

- `query_string = urlencode(params)` is computed once.
- New guard: `if end_session_endpoint.endswith(("?", "&")):` appends the query
  string with NO separator, so `"logout?" + "state=s"` → `"logout?state=s"`.
- Otherwise the original `"&" if parsed.query else "?"` logic runs unchanged.
- Regression test `test_trailing_question_mark_not_doubled` asserts `"??" not in url`
  and that `state` parses back correctly. Passes.

### Path walk of the fixed function (fix diff only)

| Path | Input | Result | Verdict |
|------|-------|--------|---------|
| empty params early-return | no params supplied | returns bare endpoint (unchanged, before new code) | OK |
| trailing `?` | `logout?` + state | `logout?state=s` | OK (fixed) |
| trailing `&` | `logout?a=1&` + state | `logout?a=1&state=s` (parsed.query=`a=1`, guard fires, direct append) | OK |
| normal endpoint, no query | `https://op/logout` + state | `?` separator → `logout?state=s` | OK |
| endpoint with existing query | `logout?foo=bar` + state | `&` separator → `logout?foo=bar&state=s` | OK |
| fragment + trailing `?` | `logout?#frag` | fragment stripped first → `logout?` → guard fires → `logout?state=s` | OK |
| degenerate `logout?&` | trailing `&`, empty parsed.query | guard fires → `logout?&state=s` | OK (better than old `logout?&?state=s`) |

### Findings

No unhandled edge cases found.

Notes considered and dismissed:
- A bare `"&"` with no preceding `"?"` (e.g. `"logout&"`) yields `"logout&state=s"`
  (no `?`). This is a malformed OP endpoint and is NOT a regression the fix
  introduced — the old code produced the equally-malformed `"logout&?state=s"`.
  Not in scope, not newly broken.
- The `conformance/app.py` log change (log endpoint instead of full URL) reads
  `logout_context["end_session_endpoint"]`, a key set earlier in the same builder;
  no new None/KeyError path introduced by the diff.

### Summary
- Unhandled paths found: 0

jamescrowley321:
## Review artifact: sentinel

None.

### WARN (should fix)
None.

### INFO (acceptable risk)
- `conformance/app.py:998` — The full end-session URL, including the `id_token_hint`
  (a complete ID Token JWT with subject/PII claims), is written to the info log
  via `logger.info("Redirecting to OP end-session endpoint: %s", end_session_url)`.
  ID Tokens are hints, not bearer credentials, and this is the conformance harness
  (test tooling, not production middleware), so exposure is limited. Worth keeping
  in mind if this logging pattern is copied into a real RP: redact `id_token_hint`
  before logging there.
- `conformance/app.py:988,1011` — `_expected_logout_state` is a module-global set on
  each `/logout` and never cleared after a successful `/post-logout-callback`
  validation. The state is therefore replayable (the same value validates on
  repeated callback hits) and a second concurrent logout would overwrite it.
  Acceptable for a single-flow, sequential conformance harness; not a pattern to
  reuse for a multi-user production RP, where state should be per-session and
  single-use.
- `conformance/app.py:966` — `/logout` is a GET with no anti-CSRF token of its own,
  so it is theoretically forgeable (login/logout CSRF). Impact is limited to
  forcing a benign logout in a test harness; acceptable here.

### Analysis notes (verified, no finding)
- State round-trip (`core/logout_logic.py:394`) is correct: empty/missing
  `expected_state` raises before any comparison, empty/missing `returned_state`
  raises next, and the final check uses `hmac.compare_digest` on two
  guaranteed-non-empty strings — mirrors `validate_authorize_callback_state`.
  No empty/missing-state bypass exists.
- `build_end_session_url` (`core/logout_logic.py:314`) routes every value through
  `urlencode`, strips URL fragments, and rejects `**extra_params` that collide
  with reserved logout params — no parameter/query injection into the
  end-session URL.
- `post_logout_redirect_uri` is fixed from the `RP_BASE_URL` env var, not
  attacker-controlled; the library intentionally does not validate it (registered-
  URI enforcement is the OP's responsibility). No RP-side open redirect.
- The 400 error page (`conformance/app.py:1013`) renders only static library
  message text and applies `html.escape`. No reflected XSS.

### Summary
- BLOCK: 0 | WARN: 0 | INFO: 3
- Overall: PASS

jamescrowley321:
## Review artifact: viper

## Review: Red Team (Viper)

### Attack Surface
- `conformance/app.py:/logout` (GET) — new unauthenticated endpoint that builds an end-session URL from the *last* completed login context and 302-redirects the user agent to the OP. Carries no request parameters; reads process-global `_last_logout_context`; writes process-global `_expected_logout_state`; logs the full redirect URL.
- `conformance/app.py:/post-logout-callback` (GET) — new endpoint that validates the OP-returned `state` against process-global `_expected_logout_state` via `validate_post_logout_state`. Inert response (static HTML), no session lookup, no state clearing.
- `core/logout_logic.py:build_end_session_url` — pure URL builder (library). Percent-encodes all params via `urlencode`, blocks reserved-param override through `**extra_params`, strips fragments, validates non-empty endpoint.
- `core/logout_logic.py:validate_post_logout_state` — pure `state` comparison (library). Uses `hmac.compare_digest` (constant-time), rejects empty/missing expected and returned values.
- `core/models.py` / `core/response_processors.py` — adds `end_session_endpoint` to discovery parsing (trusted issuer data).

### Findings

#### MEDIUM — Full ID Token (`id_token_hint`) written to server logs and browser-side history
- **Location**: `conformance/app.py:998` (`logger.info("Redirecting to OP end-session endpoint: %s", end_session_url)`); the URL is assembled at `app.py:990-996` with `id_token_hint=logout_context["id_token"]`.
- **Attack scenario**:
  1. A victim completes a normal login; `/callback` stores the issued ID Token in `_last_logout_context["id_token"]`.
  2. The victim (or the conformance runner) hits `/logout`. `build_end_session_url` places the entire signed ID Token in the `id_token_hint` query parameter.
  3. `logger.info` emits the *complete* end-session URL — including the ID Token — at INFO level to stdout/log aggregation. The same token also lands in the browser address bar / history and, on the OP side, in the `Referer` of any subresource the OP logout page loads.
  4. An attacker with read access to logs (shared CI log store, log-forwarding pipeline, or the container's stdout) harvests the ID Token, recovering its identity claims (sub, email, name — PII) and a reusable `id_token_hint` for that session.
- **Prerequisites**: Read access to application logs or the victim's browser history/Referer chain. No auth to the app required for the log path if logs are exposed.
- **CVSS v3.1**: 5.3 (AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:N/A:N) — confidentiality of identity claims; high attack complexity because it depends on log/history exposure.
- **Remediation**: Do not log the assembled end-session URL at INFO. Log only the endpoint host/path (`urlparse(end_session_url).netloc + path`) or redact query params. Treat ID Tokens as secrets in all log statements. This is a harness issue, but the pattern will be copied into the planned `fastapi-identity-model` middleware, where it becomes production-grade PII/token leakage — fix it here.

#### LOW — Post-logout `state` is a single process-global, not bound to the session that initiated logout
- **Location**: `conformance/app.py:987-988` (write of `_expected_logout_state`), `app.py:1011` (validation against the same global), declared `app.py:229` / diff `app.py:99`.
- **Attack scenario**:
  1. `_expected_logout_state` is one module-level string shared across every user/session in the process, and `/post-logout-callback` performs no session-cookie lookup — it compares the returned `state` only against this global.
  2. Victim A calls `/logout`; global is set to `stateA`.
  3. Before A's OP round-trip returns, attacker (or a second concurrent user) calls the unauthenticated GET `/logout`; the global is overwritten with `stateB` (last-writer-wins; sync endpoints run concurrently in Starlette's threadpool, so this is a genuine race).
  4. A's callback with `stateA` now fails validation (benign self-DoS of the legitimate flow), while whoever holds `stateB` passes. The check therefore does not actually bind the post-logout redirect to the browser/session that started the logout — the CSRF property the `state` round-trip is meant to provide is only nominal in any multi-session run.
- **Prerequisites**: Two concurrent logout flows in the same process (or an attacker able to reach the unauthenticated `/logout`). Impact is limited because `/post-logout-callback` is inert (returns static HTML, mutates no session/auth state).
- **CVSS v3.1**: 3.7 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N).
- **Remediation**: Store the expected `state` per-session (keyed by a logout session cookie), the same way `sessions[state]` already keys login flows, and validate the returned `state` against that session entry rather than a shared global. Critical to fix before this pattern is lifted into framework middleware.

#### LOW — `_expected_logout_state` is never cleared after use (stale/replayable state)
- **Location**: `conformance/app.py:1010-1026` — `post_logout_callback` never resets `_expected_logout_state` to `None` on success.
- **Attack scenario**:
  1. A logout completes; `_expected_logout_state` retains the last value indefinitely.
  2. Anyone who learns that `state` (via the log-disclosure path above, Referer, or history) can re-invoke `/post-logout-callback?state=<value>` and receive a 200 "Logout Successful" response repeatedly — the state is single-use by spec intent but is accepted unboundedly here.
- **Prerequisites**: Knowledge of a previously used `state`. Impact is minimal because the endpoint performs no security-sensitive action.
- **CVSS v3.1**: 2.0 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N).
- **Remediation**: Set `_expected_logout_state = None` (or delete the per-session entry once fix #2 lands) immediately after a successful or failed validation so each `state` is strictly single-use.

### Checked, not exploitable
- **Open redirect via `post_logout_redirect_uri`**: `post_logout_redirect_uri` is hard-coded to `f"{RP_BASE_URL}/post-logout-callback"` (`app.py:994`, `RP_BASE_URL` from env at `app.py:185`); it is never taken from a request parameter, so no attacker-controlled redirect target exists in the RP. `build_end_session_url` intentionally does not restrict the value (that is the OP's registration check), which is correct library behaviour.
- **Query-parameter / URL injection via `id_token_hint`, `state`, `client_id`**: `build_end_session_url` routes every value through `urlencode` (`logout_logic.py:391`), so `&`/`=`/`#` in a value cannot inject additional end-session parameters. Reserved parameters cannot be overridden through `**extra_params` (`logout_logic.py:359-364`). Fragments are stripped before appending the query.
- **Reflected XSS on `/post-logout-callback`**: the only interpolated value in the error HTML is the library's fixed `exc.message`, wrapped in `html.escape` (`app.py:1017`); the attacker-controlled `state` is never reflected into the response.
- **Timing side-channel on `state` compare**: `validate_post_logout_state` uses `hmac.compare_digest` (constant-time) and rejects empty/`None` on both sides before comparing.

### Summary
- Attack surface elements: 5
- Overall: PASS (harness-scoped; the log-disclosure and non-session-bound-state patterns MUST be fixed before being lifted into the planned framework middleware packages)

jamescrowley321:
## Review artifact: viper-delta

## Review: Red Team (Viper) — DELTA re-review

Scope: fix diff only (`/tmp/pim-RPL/.claude/review-delta.patch`). Three files:
`conformance/app.py` (log-line fix), `src/py_identity_model/core/logout_logic.py`
(bare-separator guard), `src/tests/unit/test_end_session_url.py` (regression test).

### Attack Surface
- `conformance/app.py:/logout` — INFO log line changed from the full end-session
  URL to the bare endpoint. This is the surface the prior MEDIUM was raised on.
- `logout_logic.build_end_session_url()` — new branch that appends the query
  string without a separator when the endpoint already ends in `?`/`&`.

### Verification of the addressed MEDIUM

**CONFIRMED FIXED.** The log now emits `logout_context["end_session_endpoint"]`
instead of `end_session_url`.

- Provenance of the logged value: `end_session_endpoint` is set at
  `conformance/app.py:810` from `disco.end_session_endpoint` (the OP's
  discovery-published end-session URL) and stored into `_last_logout_context`
  at 811-816. It is a bare public OP endpoint — no `id_token_hint`, no
  `state`, no `client_secret`, no bearer/refresh material. Logging it at INFO
  discloses nothing an attacker with log access couldn't already read from the
  OP's `.well-known` document.
- The full `end_session_url` (which embeds the ID Token via `id_token_hint`)
  is still built (990) and used only in the `RedirectResponse` (1004) — sending
  it to the OP is the intended RP-Initiated Logout flow, not a leak. No other
  log site emits it (grep of `end_session_url` / `logger.*id_token` clean; the
  only other token-adjacent log, 731, emits `exc.message`, not a token).
- No new leak introduced.

### Verification of the library guard (injection / param-smuggling)

**No exploitable vector.** All parameter *values* (`id_token_hint`, `state`,
`client_id`, `post_logout_redirect_uri`, `extra_params`) are passed through
`urlencode(params)` in **both** branches, so `&`, `=`, and `?` in any value are
percent-encoded. The new branch changes only whether a *separator character* is
inserted before an already-encoded query string; it cannot introduce an
unescaped delimiter, so no attacker-supplied value can smuggle an extra
parameter or break out of the query component. Reserved-param collision guard
(164-169) and fragment stripping (171-175) are untouched and still run first.

### Notes (informational, not findings)

- The guard groups `?` and `&` together, but a trailing bare `&` with *no*
  preceding `?` (e.g. `https://op/logout&`) does not actually open a query
  component; the new code would emit `.../logout&state=s`, leaving `state` in
  the path rather than the query. Impact is fail-safe (params silently dropped,
  nothing leaked or smuggled) and the input is a nonsensical OP discovery
  endpoint. Correctness edge case at most — **not a security issue**, no
  severity assigned.
- The two prior LOW items (`_expected_logout_state` process-global, not
  session-bound, never cleared) remain unaddressed by design — harness-scoped,
  serially cert-driven, consistent with the already-cleared `_last_flow_context`
  precedent. Nothing in this diff worsens them.

### Summary
- Attack surface elements: 2
- New MEDIUM+ regressions in the fix diff: none
- Overall: PASS

PR #453, opened 2026-07-21
test(keycloak): live logout + dynamic registration integration tests

jamescrowley321:
## Summary

Live, provider-agnostic + capability-gated integration tests covering the three KC.5 acceptance criteria (run where the capability exists, skip cleanly elsewhere):

- **End-session round-trip** (`test_rp_initiated_logout.py`, LOGOUT-001/002/003 live) — `build_end_session_url` with `id_token_hint` + `post_logout_redirect_uri` + `state`; follows the redirect, asserts it lands on the post-logout URI with `state` echoed, then `validate_post_logout_state`. Negative test asserts a tampered state raises `LogoutStateValidationException`. Skips cleanly if the provider interposes an interactive logout confirmation.
- **Dynamic registration CRUD** (`test_dynamic_registration_live.py`, RFC 7591/7592 live, sync + async) — `register → read → update → delete`, minting a Keycloak initial-access-token via the admin REST API (`clients-initial-access`) with `env > admin > anonymous` precedence. Threads the rotated `registration_access_token` forward across management calls (Keycloak rotates per RFC 7592 §3). Skips cleanly on 401/403 (protected registration).
- **Real logout-token validation** (`test_backchannel_logout_live.py`, LOGOUT-004..010 live) — opt-in, receiver-gated best-effort capture of a real logout token via a threaded host HTTP receiver; `validate_logout_token` asserts claims + `events` membership. Skips unless `TEST_BACKCHANNEL_LOGOUT_RECEIVER_URL` is set (shipped fixture binds loopback-only; container→host wiring is KC.6). Library correctness is already unit-covered by BCL.

Supporting wiring:
- `conftest.py` — capability detection for `end_session` / `registration` (endpoint probes) + `backchannel_logout` (feature probe); new `logout_id_token` session fixture.
- `test_utils.py` — added the four new `TEST_*` keys to the curated `get_config` allowlist (without them the gated tests silently skip).
- `.env.keycloak` — admin vars for minting initial-access-tokens + optional receiver URL.

Refs KC.5.

## Review Findings Addressed

Reviewers: Blind Hunter + Edge Case Hunter + Acceptance Auditor (KC.x override; tests-only diff).

- **Acceptance Auditor:** 3/3 ACs PASS, no scope creep. AC-1 live-token skips in the shipped loopback config (KC.6 wires container→host) = spec-permitted.
- **Edge Case Hunter (4 [CRASH]):** wrapped RP-logout follow-redirect GETs in `_drive_logout` (`try/except httpx.RequestError → skip`); swallowed the back-channel logout-GET `RequestError` (push fires before redirect; poll decides); guarded receiver `bind` (`try/except OSError → skip`).
- **Blind Hunter (5 SHOULD):** sync + async registration CRUD wrapped in `try/finally` best-effort delete (no orphan clients on a live provider); receiver `server_close()` after `shutdown()` (releases port); Content-Length parse guard; returned-state assertion.
- **Skipped w/ rationale:** committed `admin/admin` in `.env.keycloak` is pre-existing KC.1 local-only fixture creds (not KC.5-introduced), loopback-bound.
- Delta re-review (Blind + Edge, scoped to the fix diff): 0 MUST / 0 SHOULD, 0 regressions.

## Test plan
- [x] Unit tests pass (1303 passed, 95.48% cov)
- [x] Integration tests pass (node-oidc 117 passed / 10 skipped; Keycloak 97 passed / 30 skipped — all 4 KC.5 tests RUN + PASS live vs Keycloak; back-channel stays a clean opt-in skip)
- [x] Lint passes (ruff + format + pyrefly + coverage)
- [ ] CI passes

### DIFF
--- a/.env.keycloak
+++ b/.env.keycloak
@@ -10,3 +10,17 @@ TEST_AUTH_CODE_CLIENT_SECRET=test-auth-code-secret
 TEST_AUTH_CODE_REDIRECT_URI=http://localhost:8080/callback
 TEST_PKCE_PUBLIC_CLIENT_ID=test-pkce-public
 TEST_PKCE_PUBLIC_REDIRECT_URI=http://localhost:8080/callback
+# Admin credentials for minting a client-registration initial access token
+# via the Keycloak admin REST API (dynamic registration CRUD test). Bootstrap
+# admin lives in the master realm (KC_BOOTSTRAP_ADMIN_USERNAME/PASSWORD).
+TEST_ADMIN_USERNAME=admin
+TEST_ADMIN_PASSWORD=admin
+TEST_ADMIN_REALM=master
+TEST_PROVIDER_REALM=py-identity-model
+# Optional: a provider-reachable URL that captures the pushed logout_token for
+# the live back-channel logout test. The shipped fixture binds loopback-only
+# (no container->host route) so leave this unset to skip that test cleanly.
+# TEST_BACKCHANNEL_LOGOUT_RECEIVER_URL=
+# Optional: a pre-issued client-registration initial access token. When set it
+# takes precedence over minting one via the admin REST API above.
+# TEST_REGISTRATION_INITIAL_ACCESS_TOKEN=
--- a/src/tests/integration/conftest.py
+++ b/src/tests/integration/conftest.py
@@ -218,6 +218,8 @@ def _detect_grant_capabilities(raw_discovery: dict, grants: set) -> set[str]:
         ("userinfo_endpoint", "userinfo"),
         ("device_authorization_endpoint", "device_authorization_endpoint"),
         ("pushed_authorization_request_endpoint", "par"),
+        ("end_session_endpoint", "end_session"),
+        ("registration_endpoint", "registration"),
     ):
         if raw_discovery.get(endpoint):
             caps.add(cap)
@@ -234,6 +236,8 @@ def _detect_feature_capabilities(raw_discovery: dict) -> set[str]:
         caps.add("dpop")
     if raw_discovery.get("request_parameter_supported"):
         caps.add("jar")
+    if raw_discovery.get("backchannel_logout_supported"):
+        caps.add("backchannel_logout")
 
     # devInteractions: only local fixtures support automated
     # browser-like auth code flows
@@ -980,6 +984,21 @@ def auth_code_result(provider_capabilities, discovery_document, test_config):
             resource="urn:test:api",
         ),
     )
+
+
+@pytest.fixture(scope="session")
+def logout_id_token(auth_code_result):
+    """The ID Token minted by the confidential auth-code flow.
+
+    Used as the ``id_token_hint`` for RP-Initiated Logout end-session
+    round-trips. Skips cleanly if the provider issued no ID Token (e.g.
+    the ``openid`` scope was not honoured).
+    """
+    token_response = auth_code_result["token_response"]
+    id_token = (token_response.token or {}).get("id_token")
+    if not id_token:
+        pytest.skip("Auth-code token response carried no id_token")
+    return id_token
 
 
 @pytest.fixture(scope="session")
--- a/src/tests/integration/test_backchannel_logout_live.py
+++ b/src/tests/integration/test_backchannel_logout_live.py
@@ -1,0 +1,188 @@
+"""Live integration test for OpenID Connect Back-Channel Logout 1.0.
+
+Confirms the library validates a *real* logout token pushed by a running OP.
+The flow: register a client whose ``backchannel_logout_uri`` points at a local
+HTTP receiver, complete an auth-code login, trigger RP-Initiated Logout, and
+capture the ``logout_token`` the OP pushes to the receiver — then validate it
+with ``validate_logout_token``.
+
+This end-to-end push requires the OP to reach a host-side receiver. The shipped
+fixtures bind loopback-only (no container->host route), so the test is opt-in:
+it runs only when ``TEST_BACKCHANNEL_LOGOUT_RECEIVER_URL`` names a
+provider-reachable URL whose port this receiver binds. Otherwise it skips
+cleanly. Full container->host wiring is deferred to KC.6; the structural
+validation rules are already covered by the unit suite (LOGOUT-004..010).
+"""
+
+from http.server import BaseHTTPRequestHandler, HTTPServer
+import threading
+import time
+from urllib.parse import parse_qs, urlparse
+
+import httpx
+import pytest
+
+from py_identity_model import (
+    ClientDeleteRequest,
+    ClientRegistrationRequest,
+    TokenValidationConfig,
+    build_end_session_url,
+    delete_client,
+    register_client,
+    validate_logout_token,
+)
+from py_identity_model.core.logout_logic import BACKCHANNEL_LOGOUT_EVENT
+
+from .conftest import AuthCodeFlowConfig, perform_auth_code_flow
+from .test_dynamic_registration_live import _obtain_initial_access_token
+
+
+CAPTURE_TIMEOUT_SECONDS = 15.0
+POLL_INTERVAL_SECONDS = 0.25
+DEFAULT_HTTP_PORT = 80
+
+
+class _LogoutTokenReceiver(BaseHTTPRequestHandler):
+    """Captures the ``logout_token`` from an OP's back-channel POST."""
+
+    def do_POST(self):  # http.server dispatch name
+        length = int(self.headers.get("Content-Length", 0))
+        body = self.rfile.read(length).decode() if length else ""
+        token = parse_qs(body).get("logout_token", [None])[0]
+        if token:
+            self.server.captured_tokens.append(token)  # type: ignore[attr-defined]
+        self.send_response(200)
+        self.end_headers()
+
+    def log_message(self, format, *args):  # noqa: A002, ARG002  silence logging
+        return
+
+
+def _start_receiver(port: int) -> HTTPServer:
+    server = HTTPServer(("0.0.0.0", port), _LogoutTokenReceiver)  # noqa: S104
+    server.captured_tokens = []  # type: ignore[attr-defined]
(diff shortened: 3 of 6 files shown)


jamescrowley321:
## Review: Acceptance Auditor

### PASS
- [AC-1] Validate a real logout token against a live back-channel-logout provider — implemented at `src/tests/integration/test_backchannel_logout_live.py:145` (`test_validate_pushed_logout_token`). Registers a client with `backchannel_logout_uri` pointing at a local `_LogoutTokenReceiver` (`:117`), completes an auth-code login via `perform_auth_code_flow`, triggers RP-Initiated Logout with `build_end_session_url` (`:215`), captures the OP-pushed `logout_token` (`:223`), and validates it with `validate_logout_token` (`:230`), asserting the `BACKCHANNEL_LOGOUT_EVENT` member (`:240`) and `sub`/`sid` (`:241`). Capability-gated on `backchannel_logout` (conftest.py:239-240) and `registration`, and skips cleanly when `TEST_BACKCHANNEL_LOGOUT_RECEIVER_URL` is unset or no token is pushed within the timeout (`:164`, `:225`). The library validation itself is confirmed against a genuine OP-minted token. Tested at `test_backchannel_logout_live.py:145` (tier: integration/e2e).
  - Note (non-blocking): the shipped fixtures bind loopback-only, so in the default config this test always skips (live-token path exercised only when a provider-reachable receiver is wired; deferred to KC.6). The spec explicitly permits a clean skip where provider→receiver reachability is not wired, so this is compliant.

- [AC-2] End-session URL round-trip with state echo + tampered-state negative case — implemented at `src/tests/integration/test_rp_initiated_logout.py:588` (`test_end_session_state_round_trip`) and `:638` (`test_tampered_state_is_rejected`). Builds the end-session URL from discovery with `build_end_session_url` (`:609`), drives the live OP following redirects (`:617`), asserts it lands on the post-logout-redirect URI (`_matches`, `:628`/`:683`) with `state` echoed, and validates the returned state via `validate_post_logout_state` (`:636`). Skips cleanly when the provider interposes an interactive/confirmation page or rejects the sessionless request (`:629`, `:670`). Negative case asserts `LogoutStateValidationException` is raised for a mismatched stored state (`:679-680`). Capability-gated on `end_session` (`_require_end_session`, `:579`; conftest.py:221). Tested at `test_rp_initiated_logout.py:588,638` (tier: integration/e2e).

- [AC-3] Register → read → update → delete (RFC 7591/7592) with sync + async parity — implemented at `src/tests/integration/test_dynamic_registration_live.py:142` (sync `TestDynamicRegistrationLifecycle.test_register_read_update_delete`) and `:218` (async `TestDynamicRegistrationLifecycleAsync.test_register_read_update_delete`). Sync path registers (`register_client`, `:124`), reads back and asserts `client_id` match (`:163`), updates the client name and asserts it (`:177`), deletes (`:195`), and confirms the client is no longer readable (`:204`). Async path mirrors register→read→update→delete via `aio_register_client`/`aio_read_client`/`aio_update_client`/`aio_delete_client` (`:229-278`). Capability-gated on `registration` (`_require_registration`, `:51`; conftest.py:222) and skips cleanly when the provider refuses registration (`:133`, `:238`). Registration-token acquisition is provider-agnostic: explicit `TEST_REGISTRATION_INITIAL_ACCESS_TOKEN` takes precedence, with a graceful Keycloak-admin-REST fallback that returns `None` (→ anonymous registration → skip) when not configured (`:75-118`). No Keycloak-only imports in any of the three files. Tested at `test_dynamic_registration_live.py:142,218` (tier: integration/e2e).

### FAIL
_(none)_

### PARTIAL
_(none)_

### SCOPE CREEP
_(none)_ — All additions trace to an AC: conftest.py capability detection for `end_session`/`registration` (`:221-222`) and `backchannel_logout` (`:239-240`) support gating; the `logout_id_token` session fixture (`:986-1002`) supplies the `id_token_hint` for AC-2; `test_utils.py` (`:700-715`) and `.env.keycloak` env additions supply the credentials for AC-1/AC-3.

### Notes on Rigor
- All library symbols used by the tests exist and signatures match: `build_end_session_url`, `validate_post_logout_state`, `validate_logout_token`, `LogoutStateValidationException`, `register_client`/`read_client`/`update_client`/`delete_client` and their `aio` counterparts, and the `Client*Request/Response` models (verified in `src/py_identity_model/__init__.py`, `core/logout_logic.py`, `sync/logout.py`, `core/models.py`, `aio/registration.py`).
- `.error` is only accessed on failure paths (guarded by `is_successful` checks), so it does not trip the `BaseResponse` field-access guard.
- Style is consistent with existing capability-gated integration tests (`test_revocation.py`: `_require_*` helper + `@pytest.mark.integration` + clean `pytest.skip`).

### Summary
- Total ACs: 3
- Pass: 3 | Fail: 0 | Partial: 0

jamescrowley321:
## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/tests/integration/test_dynamic_registration_live.py:418-467`] The sync lifecycle registers a real client on the live provider, then runs a chain of bare `assert` statements (read/update/delete) with no `try/finally`. If any assertion fails between `_register` and `delete_client`, the registered client is never deleted and is leaked on the provider — orphaned clients accumulate across runs and can collide with `client_name` reuse on the next run. Wrap the read/update/delete in a `finally` that deletes the client.
- [`src/tests/integration/test_dynamic_registration_live.py:495-544`] Same leak in the async lifecycle: `aio_register_client` creates a client, then assertions run with no cleanup guard. A failed `assert read.is_successful` (or update) leaves the client registered on the provider with no deregistration.
- [`src/tests/integration/test_backchannel_logout_live.py:243`] The `finally` calls `server.shutdown()` but never `server.server_close()`. `shutdown()` only stops the `serve_forever` loop; the listening socket stays open until GC, so the bound port is not released. A subsequent run (or the tampered/round-trip tests reusing the same port) can hit `OSError: Address already in use`. Add `server.server_close()` after `server.shutdown()`.
- [`.env.keycloak:12-13`] `TEST_ADMIN_USERNAME=admin` / `TEST_ADMIN_PASSWORD=admin` are committed into the repo. Even for a throwaway fixture, checked-in credentials train the wrong habit and get copy-pasted; a weak default admin password on a container that may be reachable is a real exposure if the fixture is ever bound beyond loopback.
- [`src/tests/integration/test_backchannel_logout_live.py:171`] `_start_receiver(port)` is called before the id_token/registration success is confirmed, and it binds `0.0.0.0` (all interfaces, S104-suppressed). If `port` resolves to 80 (see `DEFAULT_HTTP_PORT` fallback), binding requires root and will raise `PermissionError` outside the `try`, so the server thread failure is not the skip-path the test intends. Bind loopback or validate the port before starting.

### NITPICK
- [`src/tests/integration/test_backchannel_logout_live.py:121`] `int(self.headers.get("Content-Length", 0))` will raise `ValueError` on a malformed/non-numeric `Content-Length`, crashing the handler thread silently. Guard the conversion.
- [`src/tests/integration/test_backchannel_logout_live.py:125,257-258`] `server.captured_tokens` is appended from the HTTP handler thread and read from the polling thread with no lock. List append/read is GIL-atomic in CPython today, but this is an undocumented reliance; a `threading.Event` or lock would be explicit.
- [`src/tests/integration/test_rp_initiated_logout.py:634-636`] `returned_state` may be `None` if the provider redirected but stripped `state`; it is passed straight into `validate_post_logout_state` without an explicit not-None assertion, so a stripped-state provider yields a confusing failure rather than a clear one.
- [`src/tests/integration/test_dynamic_registration_live.py:705`] `TEST_ADMIN_REALM` is defaulted to `"master"` in `get_config`, and again defaulted to `"master"` at `test_dynamic_registration_live.py:355` — duplicate default; one is redundant.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `test_rp_initiated_logout.py:617` | OP 302-redirects post-logout to localhost:8080 which has no listener | `try:\n    resp = client.get(logout_url)\nexcept httpx.RequestError:\n    pytest.skip("post-logout redirect target unreachable")` | [CRASH] httpx.ConnectError; test errors instead of skip/pass |
| `test_rp_initiated_logout.py:665` | Same dead-port redirect in tampered-state test | wrap `client.get(logout_url)` in `try/except httpx.RequestError: pytest.skip(...)` | [CRASH] ConnectError raised before state comparison; test errors |
| `test_backchannel_logout_live.py:220` | Logout GET follows OP redirect to dead localhost:8080 port | wrap `client.get(logout_url)` in `try/except httpx.RequestError: pytest.skip(...)` | [CRASH] ConnectError propagates past cleanup; test errors |
| `test_backchannel_logout_live.py:171` | Receiver port bind fails (default 80 needs root, or port already in use) | `try:\n    server = _start_receiver(port)\nexcept OSError:\n    pytest.skip("receiver port unbindable")` | [CRASH] PermissionError/OSError raised outside try; test errors not skips |

### Summary
- Unhandled paths found: 4
- Critical (crash/data loss): 4

### Notes on verification
- Confirmed via `conftest.py:885` that `perform_auth_code_flow` uses `follow_redirects=False` and breaks at the redirect_uri without ever connecting, so no HTTP server listens on `TEST_AUTH_CODE_REDIRECT_URI` (localhost:8080/callback). The logout tests instead use `httpx.Client(follow_redirects=True)`, so when the OP completes the RP-initiated logout with a real 302 to that URI, httpx attempts a live connection to the closed port and raises `httpx.ConnectError`. The `_matches(...)` skip guard is unreachable on that path because the exception fires during redirect-following, before `resp.url` is inspected. Marked PLAUSIBLE because it requires the OP (e.g. Keycloak) to have the post-logout redirect URI registered and to actually issue the redirect; if it instead renders a confirmation/error page (200), the existing skip path is taken.
- `test_backchannel_logout_live.py:171` `_start_receiver` runs before the `try:` at line 174, so any bind failure escapes the `finally: server.shutdown()` cleanup as an uncaught error. Only reachable when the opt-in `TEST_BACKCHANNEL_LOGOUT_RECEIVER_URL` is set (default port 80 when the URL omits a port, or a port already bound).

Issue #476, opened 2026-08-01
epic: PIM FAPI 2.0 Security Profile — hardening & conformance


jamescrowley321:
Epic tracking the remaining work to take PIM's RP to a **certifiable FAPI 2.0 Security Profile** posture.

PIM already ships the FAPI2 groundwork — private_key_jwt (#213), PAR/RFC 9126, JAR request objects, RFC 9207 `iss` validation (#221), DPoP/RFC 9449 (`core/dpop.py`), and FAPI2 constants + validators (`core/fapi.py`). This epic closes the remaining gaps and proves conformance externally.

**Ralph workstream:** `identity-stack-planning` → `_bmad-output/implementation-artifacts/ralph-prompts/pim-fapi2-hardening.md`.

## Tasks

- [ ] #475 — FAPI 2.0 Security Profile conformance run + cert evidence *(T255, cert)*
- [ ] #215 — mTLS client auth + certificate-bound access tokens, RFC 8705 *(T253, feature)*
- [ ] #218 — JARM — JWT-Secured Authorization Response Mode *(T254, feature)*
- [ ] #431 — repr/eq redaction guard on `RefreshTokenResponse` + `PushedAuthorizationResponse` *(T252, security-low)*

## Already shipped — FAPI2 groundwork (context, do not re-implement)

#213 private_key_jwt · #221 RFC 9207 `iss` validation · #397 jwks-cache LRU · PAR/JAR/DPoP in `core/`

## Sequencing

The two feature tasks (#215, #218) and the low-severity #431 are mutually independent. The conformance run (#475) should land **last** so it exercises the full stack — it can pass today on the **DPoP** sender-constraining path, with mTLS (#215) later adding the alternative path. Ties into cert tracking #242 and the feature → profile matrix #471.

jamescrowley321:
Scoping gap surfaced during the 2026-08-01 code review: **RS-side DPoP proof verification + `cnf.jkt` enforcement** was not tracked separately — #215 covers only the *mTLS* (`x5t#S256`) binding. Filed as **#478** (sibling of #215).

Sequencing: #478 is **not** required for the FAPI-2 *client* conformance plan (#475 tests token acquisition on the already-shipped client-side DPoP path), but it **is** the load-bearing prerequisite for the E2E token harness (#462–#474), which asserts an RS *rejects* mismatched/stolen bound tokens. Suggest slotting #478 alongside #215 (both are RS sender-constraint enforcement) ahead of the E2E assertion work.

Issue #431, opened 2026-06-30
security: low — add repr/eq guard to RefreshTokenResponse and PushedAuthorizationResponse


jamescrowley321:
Split out from #300 (all critical/high/medium findings resolved; these two low-severity items remained).

Two response dataclasses carry sensitive token material but lack the `@dataclass(repr=False, eq=False)` guard that other sensitive response types use (e.g. `TokenIntrospectionResponse` at `core/models.py:692`), so token data can leak into debug/repr output:

- **L-3:** `RefreshTokenResponse` — `core/models.py:655` is plain `@dataclass`
- **L-4:** `PushedAuthorizationResponse` — `core/models.py:744` is plain `@dataclass`

## Acceptance
- Both classes use `@dataclass(repr=False, eq=False)` consistent with the other guarded response models.
- A unit test asserts `repr()` of each does not contain token/request_uri values.

jamescrowley321:
Tracked under the FAPI 2.0 hardening epic #476.

jamescrowley321:
Delivered. Shipped in PR #479 (`fix(models): redact secrets in response repr (#431)`), released in **v3.4.3**. `RefreshTokenResponse` and `PushedAuthorizationResponse` now redact secrets in their repr. Closing as done.

PR #479, opened 2026-08-01
fix(models): redact secrets in response repr (#431)

jamescrowley321:
## Summary
Secret-bearing `BaseResponse` subclasses were plain `@dataclass`es whose generated `__repr__`/`__eq__` (a) crashed on failed responses by reading fields through the access guard and (b) leaked replayable credentials (`refresh_token`, PAR `request_uri`, token dicts, `client_secret`, `device_code`) verbatim into `repr()`/logs.

- Add an opt-in `_secret_fields: ClassVar[frozenset[str]]` to `BaseResponse`; `__repr__` renders those fields as `[REDACTED]` (empty default → no behavior change for existing models). Reuses the existing `AuthorizeCallbackResponse` redaction style.
- Convert every remaining bare `@dataclass` `BaseResponse` subclass to `@dataclass(repr=False, eq=False)` so all 14 inherit the safe, redacting base repr/eq — zero bare `@dataclass` subclasses remain.
- Redact the two named in #431 (`RefreshTokenResponse.token`, `PushedAuthorizationResponse.request_uri`) plus sibling secret-bearing responses surfaced in review: `ClientCredentials`/`AuthorizationCode`/`Device`/`TokenExchange` token dicts, `ClientRegistrationResponse` (`client_secret`, `registration_access_token`), and `DeviceAuthorizationResponse.device_code`.
- Introspection/UserInfo `claims`/`raw` PII intentionally left visible (guarded ≠ secret; designed payload — see DEC-001).

Refs #431
Refs #476

## Review Findings Addressed
5 independent review agents (Blind, Edge, Acceptance, Sentinel, Viper) ran on the initial 2-class fix. Cross-cutting P0/P1: sibling secret-bearing responses still leaked, and `DeviceAuthorizationResponse` crashed on repr in every state. Both fixed across two review-fix iterations; final delta re-review PASS (all reviewers), programmatic audit confirms all 14 subclasses inherit the safe repr. 2 NITPICKs deferred; 1 SHOULD (auto-sync `_guarded_fields`/`_secret_fields`) declined by design (DEC-001).

## Test plan
- [x] Unit tests pass (1356 passed; +28 new redaction/eq/no-crash tests)
- [x] Integration tests pass (or `[skip-integration-tests: <reason>]` justified below)
- [ ] E2E tests pass (identity-stack only, when applicable)
- [x] Lint passes (ruff + ruff-format + pyrefly + coverage 95.49%)
- [ ] CI passes

`[skip-integration-tests]`: low-severity fix confined to dataclass `repr`/`eq` redaction — no network or protocol surface exists to exercise via integration. Per #431 scope this is unit-only.

### DIFF
--- a/src/py_identity_model/core/models.py
+++ b/src/py_identity_model/core/models.py
@@ -90,6 +90,7 @@ class BaseResponse(_GuardedResponseMixin):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset()
+    _secret_fields: ClassVar[frozenset[str]] = frozenset()
 
     is_successful: bool
     error: str | None = None
@@ -97,12 +98,15 @@ class BaseResponse(_GuardedResponseMixin):
     __hash__ = None  # type: ignore[assignment]  # mutable dataclass
 
     def __repr__(self) -> str:
-        """Safe repr that bypasses field-access guards."""
+        """Safe repr that bypasses field-access guards and redacts secrets."""
         cls_name = type(self).__name__
         parts: list[str] = []
         for f in fields(self):
             val = object.__getattribute__(self, f.name)
-            parts.append(f"{f.name}={val!r}")
+            if f.name in self._secret_fields and val is not None:
+                parts.append(f"{f.name}='[REDACTED]'")
+            else:
+                parts.append(f"{f.name}={val!r}")
         return f"{cls_name}({', '.join(parts)})"
 
     def __eq__(self, other: object) -> bool:
@@ -724,7 +728,7 @@ class RefreshTokenRequest(BaseRequest):
     private_key_jwt: PrivateKeyJwt | None = None
 
 
-@dataclass
+@dataclass(repr=False, eq=False)
 class RefreshTokenResponse(BaseResponse):
     """Response from a refresh token grant.
 
@@ -734,6 +738,7 @@ class RefreshTokenResponse(BaseResponse):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset({"token"})
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})
 
     token: dict | None = None
 
@@ -815,7 +820,7 @@ class PushedAuthorizationRequest(BaseRequest):
     private_key_jwt: PrivateKeyJwt | None = None
 
 
-@dataclass
+@dataclass(repr=False, eq=False)
 class PushedAuthorizationResponse(BaseResponse):
     """Response from a pushed authorization request endpoint (RFC 9126).
 
@@ -824,6 +829,7 @@ class PushedAuthorizationResponse(BaseResponse):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset({"request_uri", "expires_in"})
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"request_uri"})
 
     request_uri: str | None = None
     expires_in: int | None = None
--- a/src/py_identity_model/core/models.py
+++ b/src/py_identity_model/core/models.py
@@ -859,7 +859,7 @@ class DeviceAuthorizationRequest(BaseRequest):
     private_key_jwt: PrivateKeyJwt | None = None
 
 
-@dataclass
+@dataclass(repr=False, eq=False)
 class DeviceAuthorizationResponse(BaseResponse):
     """Response from the device authorization endpoint (RFC 8628).
 
@@ -878,6 +878,10 @@ class DeviceAuthorizationResponse(BaseResponse):
             "interval",
         }
     )
+    # ``device_code`` is the RFC 8628 polling credential; ``user_code`` /
+    # ``verification_uri`` are meant to be shown to the user, so only the
+    # device_code is redacted.
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"device_code"})
 
     device_code: str | None = None
     user_code: str | None = None
--- a/src/py_identity_model/core/models.py
+++ b/src/py_identity_model/core/models.py
@@ -658,6 +658,7 @@ class ClientCredentialsTokenResponse(BaseResponse):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset({"token"})
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})
 
     token: dict | None = None
 
@@ -700,6 +701,7 @@ class AuthorizationCodeTokenResponse(BaseResponse):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset({"token"})
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})
 
     token: dict | None = None
 
@@ -902,7 +904,7 @@ class DeviceTokenRequest(BaseRequest):
     private_key_jwt: PrivateKeyJwt | None = None
 
 
-@dataclass
+@dataclass(repr=False, eq=False)
 class DeviceTokenResponse(BaseResponse):
     """Response from a device token poll (RFC 8628).
 
@@ -919,6 +921,7 @@ class DeviceTokenResponse(BaseResponse):
     """
 
     _guarded_fields: ClassVar[frozenset[str]] = frozenset({"token"})
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})
 
     token: dict | None = None
     error_code: str | None = None
@@ -963,7 +966,7 @@ class TokenExchangeRequest(BaseRequest):
     private_key_jwt: PrivateKeyJwt | None = None
 
 
-@dataclass
+@dataclass(repr=False, eq=False)
 class TokenExchangeResponse(BaseResponse):
     """Response from a token exchange request (RFC 8693).
 
@@ -975,6 +978,7 @@ class TokenExchangeResponse(BaseResponse):
     _guarded_fields: ClassVar[frozenset[str]] = frozenset(
         {"token", "issued_token_type"}
     )
+    _secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})
 
     token: dict | None = None
     issued_token_type: str | None = None
@@ -1165,6 +1169,9 @@ class ClientRegistrationResponse(BaseResponse):
             "registration_access_token",
             "registration_client_uri",
         }
+    )
+    _secret_fields: ClassVar[frozenset[str]] = frozenset(
+        {"client_secret", "registration_access_token"}
     )
 
     client_id: str | None = None
--- a/src/tests/unit/test_auth_code_token.py
+++ b/src/tests/unit/test_auth_code_token.py
@@ -193,3 +193,58 @@ class TestAsyncAuthCodeTokenExchange:
         response = await request_authorization_code_token_async(request)
 
         assert response.is_successful is False
+
+
+AC_SECRET_TOKEN = {
+    "access_token": "AC_SECRET_AT",
+    "refresh_token": "AC_SECRET_RT",
+    "id_token": "AC_SECRET_IDT",
+}
+
+
+@pytest.mark.unit
+class TestAuthCodeTokenResponseReprRedaction:
+    """#431: the ``token`` dict is a secret and must not leak via repr/str."""
+
+    def test_repr_redacts_token_value(self):
+        response = AuthorizationCodeTokenResponse(
+            is_successful=True, token=AC_SECRET_TOKEN
+        )
+
+        rendered = repr(response)
+        assert "AuthorizationCodeTokenResponse" in rendered
+        assert "[REDACTED]" in rendered
+        assert "AC_SECRET_AT" not in rendered
+        assert "AC_SECRET_RT" not in rendered
+        assert "AC_SECRET_IDT" not in rendered
+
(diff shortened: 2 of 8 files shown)


jamescrowley321:
## Review agent: acceptance

## Review: Acceptance Auditor

### PASS
- [AC-1] Both `RefreshTokenResponse` and `PushedAuthorizationResponse` use `@dataclass(repr=False, eq=False)` consistent with other guarded response models — implemented at `src/py_identity_model/core/models.py:731` (RefreshTokenResponse) and `src/py_identity_model/core/models.py:823` (PushedAuthorizationResponse). Verified against sibling `TokenIntrospectionResponse` at `core/models.py:770`. Tested indirectly via redaction/equality tests at `src/tests/unit/test_refresh_token.py:180` and `src/tests/unit/test_par.py:111` (tier: unit).
- [AC-2] A unit test asserts `repr()` of each does not contain token/request_uri values — the fix adds a `_secret_fields` mechanism to `BaseResponse.__repr__` (`core/models.py:100-110`) that redacts secret fields (necessary because the inherited `__repr__` otherwise prints every field value raw, so the decorator alone would not have redacted). `RefreshTokenResponse._secret_fields = {"token"}` (`core/models.py:741`); `PushedAuthorizationResponse._secret_fields = {"request_uri"}` (`core/models.py:832`). Tested at `src/tests/unit/test_refresh_token.py:153` (`test_repr_redacts_token_value`, also `test_str_redacts_token_value:162`) and `src/tests/unit/test_par.py:81` (`test_repr_redacts_request_uri`, also `test_str_redacts_request_uri:93`). All 8 new tests pass (tier: unit).

### FAIL
- (none)

### PARTIAL
- (none)

### SCOPE CREEP
- (none) — The `_secret_fields` redaction machinery added to `BaseResponse` (`core/models.py:93,106-107`) is not literally named in the spec, but it is required to satisfy AC-2: without it, `@dataclass(repr=False, eq=False)` inherits `BaseResponse.__repr__`, which printed all field values including secrets. Traceable to AC-2, not scope creep. Note: `expires_in` is deliberately left visible (asserted at `test_par.py:91`), matching the spec intent (only `request_uri` is the secret).

### Architecture Violations
- (none) — Change is a pure additive dataclass metadata + repr guard, no HTTP/DB/protocol runtime behavior; unit-test-only coverage is appropriate for this type-definition-level change (justifies no integration tests). `make lint`-relevant tests pass.

### Summary
- Total ACs: 2
- Pass: 2 | Fail: 0 | Partial: 0

jamescrowley321:
## Review agent: blind

## Review: Blind Hunter

### MUST FIX
- No findings.

### SHOULD FIX
- [`src/py_identity_model/core/models.py:36`, `src/py_identity_model/core/models.py:53`] The redacting/safe `__repr__` in `BaseResponse` only protects a subclass when that subclass is explicitly annotated `@dataclass(repr=False, eq=False)` AND populates `_secret_fields`. A subclass declared with a bare `@dataclass` (the previous state of these two classes) regenerates a default `__repr__` that overrides `BaseResponse.__repr__` and leaks every field verbatim. The protection is therefore opt-in per class and silently fails when forgotten — any other secret-bearing response type (e.g. a token-endpoint response with an `access_token`/`token` field) that still uses a plain `@dataclass` would leak secrets in logs. This fix only closes two of them.
- [`src/py_identity_model/core/models.py:24`] Redaction keys purely on `f.name in self._secret_fields`. `_guarded_fields` and `_secret_fields` are declared independently, so it is easy to add a new secret field, list it in `_guarded_fields`, and forget `_secret_fields` — the field then leaks via repr with no error or warning. The two sets should be kept in sync or derived from one another.

### NITPICK
- [`src/py_identity_model/core/models.py:25`] Non-string secrets are rendered as a quoted string literal `'[REDACTED]'`. For `RefreshTokenResponse.token` (a `dict`) the debug output shows `token='[REDACTED]'`, misrepresenting the field's actual type to anyone reading the repr.
- [`src/py_identity_model/core/models.py:24`] The `val is not None` guard means an empty-but-present secret (e.g. `token={}` or `request_uri=""`) is redacted as `[REDACTED]` while `None` is shown as-is; harmless for leakage but inconsistent, and slightly obscures debugging of malformed-but-successful responses.

jamescrowley321:
## Review agent: edge-case

## Review: Edge Case Hunter

### Findings

| `src/py_identity_model/core/models.py:654` | `repr(ClientCredentialsTokenResponse)` — `token` dict not in `_secret_fields` | `_secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})` | [DATA] access_token leaked verbatim via repr/str/logs |
| `src/py_identity_model/core/models.py:693` | `repr(AuthorizationCodeTokenResponse)` — `token` dict not in `_secret_fields` | `_secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})` | [DATA] access/refresh/id token leaked verbatim via repr/str/logs |
| `src/py_identity_model/core/models.py:770` | `repr(TokenIntrospectionResponse)` — `claims` dict not in `_secret_fields` | `_secret_fields: ClassVar[frozenset[str]] = frozenset({"claims"})` | [DATA] token subject/scope PII leaked via repr/str/logs |

### Summary
- Unhandled paths found: 3
- Critical (crash/data loss): 2

jamescrowley321:
## Review agent: sentinel

- None. The diff is a net security improvement (redacts secrets from `repr`/`str`) and introduces no new exploitable behavior. The redaction logic, the `repr=False, eq=False` switch, and the guarded-field bypass are all correct.

### WARN (should fix)
- [LIKELY] `src/py_identity_model/core/models.py:1152` (`ClientRegistrationResponse`) — Uses `@dataclass(repr=False, eq=False)` so it inherits the new `BaseResponse.__repr__`, but does **not** declare `_secret_fields`. On a successful response its long-lived credentials `client_secret` and `registration_access_token` are rendered in plaintext by `repr()`/`str()`.
  Attack scenario: application/middleware logs the response object (`logger.info("registered client %s", resp)`, `print(resp)`, or an APM tool such as Sentry capturing local variables in a traceback). The client_secret / registration access token is written to log storage, where a low-privileged operator or a log-store compromise harvests a long-lived credential.
  Mitigation: add `_secret_fields = frozenset({"client_secret", "registration_access_token"})`.

- [LIKELY] `src/py_identity_model/core/models.py:654` (`ClientCredentialsTokenResponse`) and `:694` (`AuthorizationCodeTokenResponse`) — Both use `@dataclass(repr=False, eq=False)` and inherit the new redacting `__repr__`, but neither declares `_secret_fields`, so their `token` dict (containing `access_token`, and for the auth-code case `refresh_token`/`id_token`) prints in plaintext.
  Attack scenario: same as above — any log/traceback of a successful response leaks bearer tokens.
  Mitigation: add `_secret_fields = frozenset({"token"})` to both, matching the pattern applied to `RefreshTokenResponse`.

- [LIKELY] `src/py_identity_model/core/models.py:906` (`DeviceTokenResponse`) and `:967` (`TokenExchangeResponse`) — These use plain `@dataclass` (no `repr=False`), so the dataclass-generated `__repr__` overrides `BaseResponse.__repr__` entirely; on success it reads `token` through the guard (which permits access when `is_successful`) and prints the token dict in plaintext. The new redaction never runs for these classes.
  Attack scenario: same credential-in-logs vector for device-flow and RFC 8693 token-exchange access tokens.
  Mitigation: switch both to `@dataclass(repr=False, eq=False)` and add `_secret_fields = frozenset({"token"})`.

### INFO (acceptable risk)
- `src/py_identity_model/core/models.py:106` — Redaction keys off a static `_secret_fields` set and only substitutes when the value is not `None`; a `None` secret prints as `None`, which correctly reveals no secret material. Logic is sound.
- `src/py_identity_model/core/models.py:112` — `__eq__` compares `self.__dict__` and does not go through the redacting `repr`, so equality is unaffected by redaction. The added tests confirm this. No leak.
- The two classes fixed here (`RefreshTokenResponse.token`, `PushedAuthorizationResponse.request_uri`) are correctly redacted. `request_uri` is a legitimately sensitive PAR capability handle; treating it as a secret is appropriate.

### Summary
- BLOCK: 0 | WARN: 3 | INFO: 3
- Overall: PASS

Note: the WARN findings are pre-existing gaps, not defects introduced by this diff — the change only makes them conspicuous by establishing the redaction pattern and applying it to just 2 of the ~6 secret-bearing response types. The diff is safe to merge on its own; the sibling classes should be brought into the same pattern in a follow-up (ideally the same PR, since it's a one-line-per-class change).

jamescrowley321:
## Review agent: viper

## Review: Red Team (Viper)

### Attack Surface
- `core/models.py:BaseResponse.__repr__` — new `_secret_fields` mechanism redacts declared secret fields in `repr()`/`str()`. This is a defensive control against credential leakage into logs, error trackers, and stdout.
- `RefreshTokenResponse` / `PushedAuthorizationResponse` — the only two response types the diff opts into redaction (`_secret_fields = {"token"}` / `{"request_uri"}`).
- Sibling secret-bearing response types left untouched: `AuthorizationCodeTokenResponse`, `ClientCredentialsTokenResponse`, `DeviceTokenResponse`, `TokenExchangeResponse`, `ClientRegistrationResponse`. These carry identical or more sensitive bearer material but keep the default empty `_secret_fields`, so their `repr()`/`str()` still emit the raw secret.

The redaction logic itself is sound: `__repr__` uses `object.__getattribute__` (no guard crash on failed responses), redacts only non-`None` secret values, and `__eq__`/hashing are unaffected. The exposure is a coverage gap, not a logic bug — the new control gives a false sense of "responses are safe to log" while the highest-value response still leaks.

### Findings

#### HIGH — Auth-code / device / exchange token dicts still leak full access + refresh + id tokens via repr/str
- **Location**: `src/py_identity_model/core/models.py:693` (`AuthorizationCodeTokenResponse`), `:653` (`ClientCredentialsTokenResponse`), `:905` (`DeviceTokenResponse`), `:966` (`TokenExchangeResponse`)
- **Attack scenario**:
  1. A consuming service completes a login and holds an `AuthorizationCodeTokenResponse` whose `.token` dict contains `access_token`, `refresh_token`, and `id_token`.
  2. The service logs the response object the way developers routinely do — `logger.info("token resp: %s", response)`, a Sentry breadcrumb/`extra`, an uncaught-exception frame that renders locals, or `print(response)` in a debug build. Because this class defines no `_secret_fields`, `__repr__` emits `token={'access_token': 'ey...', 'refresh_token': '...', 'id_token': 'ey...'}` in full.
  3. Because the diff just shipped redaction for `RefreshTokenResponse`, an operator reasonably assumes response objects are now log-safe and does exactly this — the partial control actively encourages the unsafe pattern.
  4. An attacker with read access to the log sink (centralized log aggregation, Sentry project, shared log files, or via a log-injection/forwarding misconfiguration) harvests the live `refresh_token` and `id_token`. The refresh token yields long-lived, renewable access → full account takeover that survives access-token expiry.
- **Prerequisites**: A consumer that logs/serializes the response object, plus attacker read access to that log/error sink. No valid account of the victim required.
- **CVSS v3.1**: 7.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N) — network-reachable log sinks, high confidentiality impact (live refresh token = full ATO), AC:H because it depends on the consumer emitting the repr.
- **Remediation**: Add `_secret_fields: ClassVar[frozenset[str]] = frozenset({"token"})` to `ClientCredentialsTokenResponse`, `AuthorizationCodeTokenResponse`, `DeviceTokenResponse`, and `TokenExchangeResponse`. Better: default the redaction at the base layer (e.g. auto-redact any guarded dict/token field) so new token-bearing responses are safe-by-default instead of opt-in, closing the whole class of gap rather than one field at a time.

#### MEDIUM — Client registration secrets (client_secret, registration_access_token) leak via repr/str
- **Location**: `src/py_identity_model/core/models.py:1151` (`ClientRegistrationResponse`)
- **Attack scenario**:
  1. A dynamic-registration flow returns `ClientRegistrationResponse` with `client_secret` and `registration_access_token` populated.
  2. The response is logged/serialized as above; `__repr__` prints both credentials in clear text because no `_secret_fields` is declared.
  3. An attacker reading the log obtains the long-lived `client_secret` (authenticate as the client at the token endpoint) and the `registration_access_token` (read/modify/delete the client registration per RFC 7592 — e.g. rewrite `redirect_uris` to an attacker-controlled callback to hijack future auth-code flows).
- **Prerequisites**: A consumer that logs the registration response + attacker read access to the sink. Impact escalates via the redirect_uri rewrite into an auth-code interception primitive.
- **CVSS v3.1**: 6.5 (AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:L/A:N) — client credential disclosure plus limited integrity impact via registration mutation.
- **Remediation**: Add `_secret_fields: ClassVar[frozenset[str]] = frozenset({"client_secret", "registration_access_token"})` to `ClientRegistrationResponse`.

### Summary
- Attack surface elements: 1 (repr redaction control) with 5 uncovered secret-bearing response types
- Findings: 0 critical, 1 high, 1 medium, 0 low
- Overall: FAIL — the redaction control is correct but applied inconsistently; the primary login response (`AuthorizationCodeTokenResponse`) and registration secrets remain loggable in clear text, and the partial fix increases the chance operators log them believing they are protected. Extend `_secret_fields` to all token-/credential-bearing responses (or make redaction default-on for guarded secret fields) before this ships.

PR #489, opened 2026-08-03
feat(security): opt-in multi-tenant issuer pinning (SC2)

jamescrowley321:
**SC2** — multi-tenant issuer pinning, the direct answer to the cross-tenant / spoofed-token concern (audit **RT-SPOOF-F1**). **Stacked on SC1 (#488).**

## Purely opt-in — default behaviour unchanged
`allowed_issuers` defaults to `None`. When `None` (the default), **discovery is authoritative and its issuer is used exactly as before** — pinning is skipped entirely. The full 1508-test suite passes untouched (no existing discovery test changed), which is the proof it's non-breaking. Discovery is assumed correct unless a caller explicitly pins.

Why a new field (not reusing `issuer`): in discovery mode the discovery issuer already *overrides* `config.issuer`, so there was no existing knob to pin against disco. Reusing `issuer` would silently change behaviour for callers who set both `issuer` + `perform_disco`; a new opt-in field is the conservative choice.

## The control
- `TokenValidationConfig.allowed_issuers` — when set, the issuer used to validate the token (discovery issuer in disco mode, else `issuer`) **must** be a member, enforced in `decode_with_config` (the shared chokepoint for disco / cached / retry / offline; sync **and** aio) *before* the discovery result is trusted. Fail-closed when configured-but-unresolvable.
- `build_resolved_config` propagates `allowed_issuers` so the disco/retry paths (which validate via the resolved config) enforce the pin too — otherwise the main attack path would silently skip it.

## Fail-closed tests (`test_issuer_pinning.py`)
Attacker-tenant issuer rejected at the chokepoint · **default (no allowlist) does NOT pin** (opt-in lock-in) · resolved-config propagation · allowlist logic (single/list/unresolvable). Control-matrix R.9 → shipped.

`make lint` green (ruff + pyrefly + 80% coverage). Refs Epic 16 R.9, #476, the audit. SC3 (RS-side sender-constraint) branches off this.


jamescrowley321:
**Holding this PR** — not a clean merge. After #488 (SC1 alg-confusion) squash-merged, this branch (`feat/issuer-allowlist`) still carries **SC0 (`102e9dd` security-gate foundation), SC1 (`3d5b419`, now redundant/conflicting vs the squashed version on main), SC2 (`4ee2418` issuer pinning — the intended scope), and SC3 (`288314e` RS cert-binding)**. As-is it conflicts with main (CONFLICTING) and would drag SC0/SC3 in beyond the stated SC2 scope. Needs manual rebase to isolate just SC2 onto `main`, resolve the SC1 conflict, and re-verify before merge. Retargeted base to `main` (original base `fix/alg-confusion-downgrade` was deleted by #488).

jamescrowley321:
Superseded. This Aug-3 SC-stack branch stacked SC1+SC2+SC3 on the gate foundation and has since diverged from main: the gate foundation, alg-confusion (SC1, via #507), and cert-binding all landed on main through other PRs, leaving this branch conflicting (DIRTY). Its one genuinely-unlanded feature — SC2 multi-tenant issuer pinning — is re-cut clean onto current main in #512, green through the now-real mutation gate. SC3 (sender-constraint/strict-audience) remains a tracked backlog item (control-matrix R.2).

PR #497, opened 2026-08-05
fix(discovery): SSRF-harden PAR + mTLS discovery endpoints (F-05, F-06)

jamescrowley321:
Stacked on #496 (the adversarial test suite). Merge #496 first, then this.

## F-06 — PAR endpoint SSRF (HIGH)
`pushed_authorization_request_endpoint` was omitted from the `_endpoint_names` authority/scheme-validation loop, so a malicious/compromised OP could point it at a foreign or plaintext host and exfiltrate `client_id`/`redirect_uri`/PKCE `code_challenge`/`client_secret` via the client's PAR POST. PAR endpoints are same-host as the AS, so the existing issuer-authority guard is correct — this just adds the field to the loop.

**Un-xfails** `test_discovery_ssrf.py::test_par_endpoint_ssrf_is_rejected` (now passes). `make lint` green; 515 unit tests pass, no regression.

## Why F-05 (mtls_endpoint_aliases) is NOT in this PR
The naive "add it to the same authority loop" fix **breaks a real, tested feature**: RFC 8705 §5 mTLS endpoints legitimately live on a **separate host** (`test_mtls.py::TestDiscoveryParsesMtlsFields` asserts `mtls.example.com` under issuer `example.com` must parse). Strict issuer-authority matching would reject legitimate cross-host mTLS. I won't weaken the F-05 proving test to force a fix. The correct approach — **scheme validation + an internal-IP-literal block** (blocks `http://`/`https://169.254…`/RFC1918 while allowing cross-host hostnames) — is a design call flagged for James; F-05's test stays `xfail` until then.

### DIFF
--- a/src/py_identity_model/core/response_processors.py
+++ b/src/py_identity_model/core/response_processors.py
@@ -158,6 +158,7 @@ def validate_and_parse_discovery_response(
         "registration_endpoint",
         "introspection_endpoint",
         "end_session_endpoint",
+        "pushed_authorization_request_endpoint",
     ]
     try:
         discovery_url = str(response.url)
--- a/src/py_identity_model/core/response_processors.py
+++ b/src/py_identity_model/core/response_processors.py
@@ -7,6 +7,7 @@ sync and async implementations.
 
 from __future__ import annotations
 
+import ipaddress
 from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


jamescrowley321:
Folded into #501. The squash-merge of #496 (its base) auto-closed this PR and GitHub won't reopen a PR whose base branch was deleted. The SSRF fixes (F-05 `5d60410`, F-06 `d9ee46c`) are now rebased onto `main` and travel with #501, which was retargeted to `main` and will merge-commit so the individual `fix(discovery)` + `feat(fastapi)` commits land intact.

PR #506, opened 2026-08-10
fix(discovery): close IP-encoding SSRF bypass + surface off-host escape hatch (F-05/F-06)

jamescrowley321:
First of the trust-rebuild remediation PRs. Test-first; every claim below is proven by a test that **fails on `main` and passes with the fix** (verified locally).

## What was wrong (shipped live in 3.8.5)
1. **IP-encoding SSRF bypass (F-05).** The internal-IP guard on `mtls_endpoint_aliases` only recognised dotted-quad literals — `ipaddress.ip_address("2130706433")` raises `ValueError`, so **decimal (`2130706433`), octal (`0177.0.0.1`), hex (`0x7f.0.0.1`)** encodings of `127.0.0.1` were treated as hostnames and **allowed**, while the OS resolver still routes them internally. Untested, real, exploitable.
2. **Off-host endpoints break silently (F-06 over-strictness).** The F-06 PAR authority-pin is correct to keep (a compromised OP must not redirect the client's `client_secret`/PKCE POST), but RFC 8414 §2 / RFC 9126 §5 permit endpoints on a **different host** than the issuer, so a conformant split-host OP was rejected by default with an error that didn't say how to allow it.

## Fix
- Parse the host through `socket.inet_aton` (matching resolver behaviour) so an internal IP in **any** encoding is rejected. A genuine hostname (`mtls.as.example.com`) and the already-covered IPv4-mapped IPv6 form are unaffected — **zero over-rejection**.
- Keep the PAR authority-pin **default-on**, but make the escape hatch discoverable: the authority-mismatch error now names `DiscoveryPolicy.additional_endpoint_base_addresses`, so a legit split-host OP is a one-line config, not a silent break.

## Tests (fail without fix → pass with)
- 3 encoding-evasion cases (decimal/octal/hex mtls alias → rejected).
- Positive control: off-host PAR **rejected by default** (error names the escape hatch) and **accepted once its base is allow-listed**.
- Existing SSRF + cross-host-mTLS-allowed + sibling-field controls unchanged. Full unit+security suite: **1532 passed / 17 xfailed**, no regression.

## Spec basis
RFC 8705 §5 (mTLS aliases foreign-host, IP-literal reject is spec-neutral), RFC 8414 §2 / RFC 9126 §5 (endpoints MAY be off-host → the allowlist escape hatch is mandatory, present).

## Out of scope (tracked)
- DNS-rebinding (a hostname that *resolves* internally) — needs resolve-and-pin at the HTTP layer (T9).
- **Mutation gate gap (separate PR):** `response_processors.py` isn't in `SECURITY_MODULES`, and the gate's meta-test aborts mutmut under its `mutants/` copy — so the SSRF surface was never mutation-covered and the gate can't currently run on a changed module. Constraint proof here is the fail-without/pass-with tests; the gate hardening (add module + coverage-scoped run + meta-test fix) will be its own PR.

### DIFF
--- a/src/py_identity_model/core/discovery_policy.py
+++ b/src/py_identity_model/core/discovery_policy.py
@@ -46,6 +46,14 @@ class DiscoveryPolicy:
             or CDN setups).
         authority: Expected authority (scheme + host) for endpoint
             validation. When ``None``, derived from the discovery URL.
+        allow_loopback_endpoints: Allow ``mtls_endpoint_aliases`` to target a
+            loopback address (``127.0.0.0/8`` / ``::1``) for local development.
+            Default ``False`` — a loopback-routed mTLS request is a limited SSRF
+            to the client's own host, so it must be opted into. Link-local
+            (cloud metadata, ``169.254.0.0/16``) and RFC1918/reserved hosts stay
+            blocked regardless of this flag. Distinct from
+            ``allow_http_on_loopback`` (which relaxes only the HTTPS scheme
+            requirement, not internal-IP routing).
     """
 
     require_https: bool = True
@@ -55,6 +63,7 @@ class DiscoveryPolicy:
     require_key_set: bool = True
     additional_endpoint_base_addresses: list[str] = field(default_factory=list)
     authority: str | None = None
+    allow_loopback_endpoints: bool = False
 
 
 @dataclass
--- a/src/py_identity_model/core/response_processors.py
+++ b/src/py_identity_model/core/response_processors.py
@@ -8,6 +8,7 @@ sync and async implementations.
 from __future__ import annotations
 
 import ipaddress
+import socket
 from typing import TYPE_CHECKING
 from urllib.parse import urlparse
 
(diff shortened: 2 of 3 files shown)


jamescrowley321:
Added per review feedback on localhost debugging: **`DiscoveryPolicy.allow_loopback_endpoints`** (default `False`).

The internal-IP reject only applies to `mtls_endpoint_aliases` (not the issuer/regular endpoints, which authority-pin), so ordinary localhost dev — issuer + endpoints on `127.0.0.1`/`localhost`, no mTLS — was already unaffected. The gap was an inconsistency: `allow_http_on_loopback=True` (default) says "loopback dev is fine" for the scheme check, yet the IP guard blocked loopback unconditionally with no override.

Now a developer can opt in to loopback-routed mTLS aliases via `DiscoveryPolicy(allow_loopback_endpoints=True)`. Secure-by-default (opt-in), and **not** gated on `allow_http_on_loopback` (which defaults True and would weaken production). **Link-local (169.254 cloud metadata) and RFC1918 stay blocked even with the flag on** — proven by `test_link_local_mtls_alias_rejected_even_with_loopback_opt_in`. `localhost` (a hostname) needs no flag.

Issue #462, opened 2026-08-01
epic: E2E Token-Blaster Harness & PIM Feature-Proof (TH-1..TH-3)


jamescrowley321:
Tracking issue for the **Token Validation Harness**, feature-proof backfill, and conformance gates.

Planning: identity-stack-planning `_bmad-output/planning-artifacts/epics-token-harness.md` + task-queue `Token Validation Harness` section.

**Why:** #461 (jwks-cache FIFO→LRU) merged with **no integration proof**. Unit-green is not the bar — PIM features require integration tests, plus conformance where applicable.

Architecture: a prod FastAPI server on the `fastapi-identity-model` middleware, exercised by a **pytest correctness matrix** + deterministic behavior proofs and a **Locust load/soak suite** across the wired IdP matrix (node-oidc, Keycloak, Ory, Descope multi-tenant).

### TH-1 — Harness Foundation
- [ ] TH-1.1 #463 — unified multi-provider `TokenSource` minter
- [ ] TH-1.2 #464 — boot `fastapi-identity-model` as a real resource server
- [ ] TH-1.3 #465 — token correctness assertion matrix (pytest)
- [ ] TH-1.4 #466 — `make test-harness` + CI gating
- [ ] TH-1.5 #474 — load/soak suite (Locust)

### TH-2 — Feature-Proof Backfill
- [ ] TH-2.1 #467 — prove multi-tenant LRU or revert #461
- [ ] TH-2.2 #468 — RFC 9207 issuer mix-up rejection
- [ ] TH-2.3 #469 — private_key_jwt end-to-end
- [ ] TH-2.4 #470 — recent-batch coverage sweep

### TH-3 — Conformance Coverage & Gates
- [ ] TH-3.1 #471 — feature→profile coverage matrix
- [ ] TH-3.2 #472 — fastapi RP conformance regression gate
- [ ] TH-3.3 #473 — RP-initiated + back-channel logout conformance

jamescrowley321:
**Harness-surfaced middleware candidate fixes filed** (from the load/soak design §9, verified against `origin/main`):

- #513 — expose `allowed_issuers` on the middleware (config field shipped in #512/3.9.0 but unreachable from the ASGI surface)
- #514 — expose `leeway` (hardcoded to 0 today → boundary-`exp` 401 flaps under soak)
- #515 — verify upstream outage returns 503 not 401 (the 503 branch exists; reachability unproven)
[...]

PR #520, opened 2026-08-13
feat(cache): per-process cache observability counters + benchmark + doc fix (T299)

jamescrowley321:
## Summary
Foundation fixups for the E2E Token-Blaster harness (design §8 P0 — unblocks T311's cache-hit-rate metric). Three deliverables:

- **Per-process cache observability counters** (`core/cache_metrics.py`): dependency-free, thread-safe `CacheCounters` singleton tracking disco/JWKS hit/miss and JWKS refresh. `get_cache_counters().snapshot()` surface; increments are O(1) under a leaf lock and never wrap I/O.
- **Instrumented the async cache path** (`aio/token_validation.py`): `_get_disco_response`, `_get_cached_jwks`, `_refresh_jwks` now record hit/miss/refresh. Miss/refresh counted **only on a successful upstream fetch** — scheme-rejected (zero-I/O) requests do not inflate the tally (fail-closed).
- **Warm-path micro-benchmark** (`src/tests/benchmarks/`): `test_bench_validate_token_warm_path` benches the true warm request path (disco hit + JWKS hit + decode), not just raw `pyjwt.decode`.
- **Corrected `docs/performance.md`**: the caching section documented `functools.lru_cache`/`async_lru.alru_cache(maxsize=128)` + `cache_info()`/`cache_clear()` APIs that no longer exist. Rewritten to the real TTL-`OrderedDict` LRU stack (64-entry LRU, 32-stripe single-flight, per-URI kid-miss cooldown, per-loop `WeakKeyDictionary` locks, `async clear_*_cache()`) plus the new counters surface.

Refs epic #462 (TH-1 foundation, design §8 P0).

## Review Findings Addressed
Reviewers: Blind + Edge + Acceptance + Sentinel + Viper.
- **P0: 0**
- **P1: 3** (Blind ×3 / Edge ×2, same root cause) — miss/refresh counters were incremented *before* the upstream fetch, so a scheme-rejected zero-I/O request inflated the miss/upstream-fetch tally. **Fixed** (commit 003724c): gate the increments on `response.is_successful`; docstring aligned; 3 fail-closed tests added (scheme-rejected disco/JWKS/refresh assert counter stays 0; positive control asserts refresh==1).
- **P2: 2** NITPICK (annotation-position `threading.Lock`; dataclass eq-by-value on a singleton) — skipped per triage (pyrefly passes; cosmetic on a process singleton).
- Acceptance: 4/4 AC PASS · Sentinel: 0 BLOCK/0 WARN (20 mutation waivers independently verified benign) · Viper: 0 findings.
- Mechanical `make security-gate`: **PASSED** — 140 mutants in changed functions all killed or waived-equivalent.

## Test plan
- [x] Unit tests pass (`test_cache_metrics.py` drives the real async cache funcs via respx; 27 targeted passed)
- [x] Security fail-closed suite passes (`test_aio_cache_fetch_controls.py`)
- [x] `[skip-integration-tests: per-process observability library fix; no real-IdP behavior to prove E2E — design note mandates "test = unit + the new micro-bench"]`
- [x] Warm-path micro-benchmark added
- [x] Lint passes (`make lint`: ruff/format/pyrefly/coverage green)
- [x] Mechanical security-gate passes (mutation testing on changed security modules)
- [x] Independent review agents passed (5 reviewers, 1 fix iteration)
- [ ] CI passes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,8 +100,9 @@ src/py_identity_model/
    - Core functions are pure (no I/O) and return result objects or raise exceptions
 
 3. **Caching Strategy**
-   - Discovery documents: Cached per-process with `functools.lru_cache` (sync) and `async_lru.alru_cache` (async)
-   - JWKS keys: Cached to avoid repeated fetches during token validation
+   - Discovery documents and JWKS are cached per-process with an explicit **TTL + LRU** cache (an `OrderedDict` bounded to `DEFAULT_MAX_CACHE_ENTRIES` = 64), **not** `functools.lru_cache`/`async_lru.alru_cache`. Cache *policy* lives in `core/jwks_cache.py`; the sync/async cache stacks live in `sync/token_validation.py` and `aio/token_validation.py`.
+   - Entries expire on a TTL resolved from `Cache-Control: max-age` → env override (`DISCO_CACHE_TTL` / `JWKS_CACHE_TTL`) → built-in default, clamped to `[60s, 86400s]`. Concurrent misses are coalesced into a single upstream fetch via per-URI striped locks.
+   - Per-process cache hit/miss/refresh counters are exposed via `core/cache_metrics.py` (`get_cache_counters()`, re-exported at the top level); the async cache paths are instrumented. Clear caches with `clear_discovery_cache()` / `clear_jwks_cache()` (async variants must be `await`ed).
    - All caches are per-process, not shared across processes
 
 4. **Error Handling**
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -10,112 +10,108 @@ py-identity-model is designed for high performance with built-in caching and sup
 
 ### Discovery Document Caching
 
-Discovery documents are cached automatically to avoid repeated HTTP requests:
+Discovery documents and JWKS are cached in-process to keep the steady-state
+request path off the network. The cache is **not** `functools.lru_cache` /
+`async_lru.alru_cache` — those were replaced (v3.0.0) with an explicit
+**TTL + LRU** cache so entries expire on a schedule, key rotation is handled
+automatically, and concurrent misses are coalesced into a single upstream
+fetch. The cache *policy* lives in `core/jwks_cache.py`; the sync and async
+cache stacks live in `sync/token_validation.py` and `aio/token_validation.py`.
 
-**Sync Implementation:**
-```python
-from functools import lru_cache
+### What is cached
 
-@lru_cache(maxsize=128)
-def _get_disco_response(disco_doc_address: str) -> DiscoveryDocumentResponse:
-    """Cached discovery document fetching"""
-    request = DiscoveryDocumentRequest(address=disco_doc_address)
-    return get_discovery_document(request)
-```
+| Cache | Cache key | Backing store | Default TTL | Env override |
+|-------|-----------|---------------|-------------|--------------|
+| Discovery document | `(address, require_https)` | `OrderedDict` | 3600s (1 hour) | `DISCO_CACHE_TTL` |
+| JWKS | `jwks_uri` | `OrderedDict` | 86400s (24 hours) | `JWKS_CACHE_TTL` |
 
-**Async Implementation:**
-```python
-from async_lru import alru_cache
+Both caches share the same policy machinery:
 
-@alru_cache(maxsize=128)
-async def _get_disco_response(disco_doc_address: str) -> DiscoveryDocumentResponse:
-    """Cached async discovery document fetching"""
-    request = DiscoveryDocumentRequest(address=disco_doc_address)
-    return await get_discovery_document(request)
-```
+- **LRU eviction, 64 entries** (`DEFAULT_MAX_CACHE_ENTRIES`). The store is an
+  `OrderedDict`, not a plain dict, so a *read hit* refreshes recency
+  (`touch_cache_entry`) and eviction targets the least-recently-**used** entry.
+- **TTL resolution order.** For each entry the TTL is the first available of:
+  (1) the response's `Cache-Control: max-age`, (2) the environment override,
+  (3) the built-in default. The resolved value is clamped to
+  **[60s, 86400s]** (`MIN_CACHE_TTL_SECONDS` … `MAX_CACHE_TTL_SECONDS`); a bad
+  or out-of-range env value is clamped and logged, never crashes the request.
+- **Single-flight fetch.** Per-URI striped locks (32 stripes) ensure a cache
+  miss issues exactly one upstream fetch; concurrent requests for the same URI
+  wait on the stripe and then read the freshly-cached entry (a
+  *double-checked* hit) rather than each hitting the network.
+- **Kid-miss cooldown.** If a JWT's `kid` is absent from the cached JWKS the
+  cache is treated as stale and a refresh is forced (OP key rotation). This is
+  rate-limited per-URI — default 5s (`DEFAULT_KID_MISS_REFRESH_COOLDOWN_SECONDS`,
+  override `KID_MISS_REFRESH_COOLDOWN`) — so an attacker forging tokens with
+  random unknown kids cannot amplify inbound traffic into upstream JWKS fetches.
+- **Per-event-loop locks (async).** The async write/fetch locks are keyed on
+  the running event loop via `WeakKeyDictionary` (#399), so a test runner or
+  embed that creates a new loop per scope does not hit
+  `RuntimeError: <Lock> is bound to a different event loop`.
 
-### JWKS Caching
-
-JSON Web Key Sets are also cached:
-
-- **Cache Size**: 128 entries (both sync and async)
-- **Cache Key**: JWKS URL (from discovery document)
-- **Cache Duration**: Lifetime of the process
-- **Thread Safety**: Both implementations are thread-safe
-
-### Cache Behavior Differences
-
-| Aspect | Synchronous | Asynchronous |
-|--------|-------------|--------------|
-| Library | `functools.lru_cache` | `async_lru.alru_cache` |
-| Max Entries | 128 | 128 |
-| Thread Safety | ✅ Yes | ✅ Yes |
-| Cache Sharing | Sync calls only | Async calls only |
-| Cache Clearing | `function.cache_clear()` | `function.cache_clear()` |
-| Performance | Fast | Fast |
-
-**Important:** Sync and async caches are **separate**. They do not share cached data.
-
-#### Example: Cache Separation
-
-```python
-from py_identity_model import get_discovery_document as sync_disco
-from py_identity_model.aio import get_discovery_document as async_disco
-from py_identity_model import DiscoveryDocumentRequest
-
-# First call - fetches from network (sync cache)
-disco = sync_disco(DiscoveryDocumentRequest(address="https://..."))
-
-# Second call - uses sync cache (fast)
-disco = sync_disco(DiscoveryDocumentRequest(address="https://..."))
-
-# Async call - fetches from network again (separate cache!)
-disco = await async_disco(DiscoveryDocumentRequest(address="https://..."))
-
-# Second async call - uses async cache (fast)
-disco = await async_disco(DiscoveryDocumentRequest(address="https://..."))
-```
+**Important:** the sync and async caches are **separate** process-global
+`OrderedDict`s. They do not share cached data.
 
 ### Clearing Caches
 
-You can manually clear caches if needed:
+Use the public cache-clear helpers. **Breaking change (v3.0.0):** the async
+helpers are now `async def` — they acquire the cache write lock before
+clearing so a fetch in flight can't write its result back into a "cleared"
+cache — so callers must `await` them.
 
 ```python
-from py_identity_model.sync.token_validation import _get_disco_response, _get_jwks_response
+# Sync
+from py_identity_model import clear_discovery_cache, clear_jwks_cache
 
-# Clear discovery document cache
-_get_disco_response.cache_clear()
-
-# Clear JWKS cache
-_get_jwks_response.cache_clear()
+clear_discovery_cache()
+clear_jwks_cache()
 ```
-
-For async:
-```python
-from py_identity_model.aio.token_validation import _get_disco_response, _get_jwks_response
-
-# Clear async discovery document cache
-_get_disco_response.cache_clear()
-
-# Clear async JWKS cache
-_get_jwks_response.cache_clear()
-```
-
-### Cache Statistics
-
-You can inspect cache performance:
 
 ```python
-from py_identity_model.sync.token_validation import _get_disco_response
+# Async — must be awaited
+from py_identity_model.aio import clear_discovery_cache, clear_jwks_cache
 
-# Get cache statistics
-info = _get_disco_response.cache_info()
-print(f"Hits: {info.hits}")
-print(f"Misses: {info.misses}")
-print(f"Size: {info.currsize}")
-print(f"Max Size: {info.maxsize}")
-print(f"Hit Rate: {info.hits / (info.hits + info.misses) * 100:.2f}%")
+await clear_discovery_cache()
+await clear_jwks_cache()
 ```
+
+There is no `_get_disco_response.cache_clear()` / `.cache_info()` — those are
+`functools.lru_cache` APIs and the cache no longer uses `lru_cache`.
+
+### Observing Cache Hit Rate
+
+The cache exposes lightweight per-process counters via
+`core/cache_metrics.py`, re-exported at the top level. Read a `snapshot()` and
+compute whatever rate you need:
+
+```python
+from py_identity_model import get_cache_counters
+
+snap = get_cache_counters().snapshot()
+# {'disco_hits': ..., 'disco_misses': ..., 'jwks_hits': ...,
+#  'jwks_misses': ..., 'jwks_refreshes': ...}
+
+jwks_total = snap["jwks_hits"] + snap["jwks_misses"]
+jwks_hit_rate = snap["jwks_hits"] / jwks_total if jwks_total else 0.0
+```
+
+Semantics:
+
+- **hit** — request served from a fresh cached entry (no upstream call);
+  **miss** — request that had to fetch from upstream;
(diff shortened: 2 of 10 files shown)


jamescrowley321:
### review-blind

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/py_identity_model/aio/token_validation.py:247`] `record_jwks_miss()` is called *before* `get_jwks(...)`, so the miss is tallied even when the request never reaches upstream. The pre-flight scheme check inside `get_jwks` can reject a plaintext `http://` JWKS URI with zero network I/O (the diff's own test `test_cached_jwks_default_rejects_plaintext_http` asserts `route.call_count == 0` while the call still passes through this line). This contradicts the counter's documented semantics in `core/cache_metrics.py` ("A *miss* is a request that had to fetch the document/keys from upstream") and lets an attacker inflate the miss count with rejected tokens, distorting the very hit-rate the feature exists to report. Increment only on a fetch that was actually issued (or record miss only after a successful/attempted upstream call).
- [`src/py_identity_model/aio/token_validation.py:291`] Same defect for `record_jwks_refresh()`: it is incremented before `get_jwks(...)`, so a refresh against a plaintext `http://` URI under the default `require_https=True` policy is counted as an "upstream re-fetch" even though `get_jwks` rejects it before any network call (see `test_refresh_default_rejects_plaintext_http`, which asserts `route.call_count == 0`). The docstring claims a refresh "coalesces ... does no upstream work and is not counted", yet a rejected refresh does no upstream work and *is* counted here.
- [`src/py_identity_model/aio/token_validation.py:154`] `record_disco_miss()` is likewise recorded before `get_discovery_document(...)` returns/raises, so a discovery fetch that raises mid-flight is still counted as a miss with no corresponding cache population. Lower impact than the JWKS paths (discovery failures generally do hit the network first), but the same "count before the work succeeds" pattern; move the increment to reflect an actually-attempted upstream fetch for consistency.

### NITPICK
- [`src/py_identity_model/core/cache_metrics.py:392`] `_lock: threading.Lock = field(default_factory=threading.Lock, ...)` uses `threading.Lock` in annotation position, where it is a factory function rather than a type. Works at runtime, but it is a well-known footgun for static type checkers; prefer no annotation with the `field(...)` default, or annotate with the actual lock type.
- [`src/py_identity_model/core/cache_metrics.py:375`] `CacheCounters` is a mutable `@dataclass` with default `eq=True`. Two independently-constructed counters compare equal whenever their tallies match (the `_lock` is `compare=False`), and the class is unhashable. Harmless given it is used as a process singleton, but equality-by-value on a counter object is a surprising contract to expose publicly (it is in `__all__`).

jamescrowley321:
### review-edge-case

## Review: Edge Case Hunter

### Findings

| `src/py_identity_model/aio/token_validation.py:247` | JWKS URI fails `get_jwks` pre-flight scheme check (e.g. `http://`, `ftp://`) | `resp = await get_jwks(...)` then `if resp.is_successful or _network_attempted: record_jwks_miss()` | [DEGRADED] miss counted though zero upstream fetch occurred; inflates upstream-fetch-volume metric |
| `src/py_identity_model/aio/token_validation.py:154` | Disco address fails `get_discovery_document` pre-flight scheme check before any network I/O | move `record_disco_miss()` to after a fetch is actually attempted | [DEGRADED] disco miss counted with no upstream fetch; overstates fetch volume |

### Summary
- Unhandled paths found: 2

### Notes (verification detail, not additional findings)

The changed production surface is small: the new `core/cache_metrics.py`
counter class and five `get_cache_counters().record_*()` call sites inserted
into `aio/token_validation.py`, plus docs/tests. `CacheCounters` is
lock-guarded, all methods are O(1) integer ops with no None/empty/overflow
paths, `snapshot()` returns a detached copy, and `get_cache_counters()` cannot
raise. No crash, data-loss, or security-control path is introduced by the diff.

The two DEGRADED findings above are the only genuinely wrong-result paths: the
`record_*_miss()` increments sit **before** the `get_jwks` /
`get_discovery_document` call, but those functions reject disallowed URL schemes
in a pre-flight check (`aio/jwks.py:59`) that raises before any network request
(the security test at `src/tests/security/test_aio_cache_fetch_controls.py`
asserts `route.call_count == 0` for exactly this path). The module docstring
defines a miss as "a request that had to fetch ... from upstream" and states the
counters make "upstream-fetch volume observable"; a scheme-rejected request
performs zero upstream work, so counting it as a miss contradicts the documented
semantics and overstates fetch volume. Impact is confined to observability
counters — no functional, caching, or security behavior is affected.

jamescrowley321:
### review-acceptance

## Review: Acceptance Auditor

Task T299 is a design-internal foundation-fixup (no GitHub issue). ACs are drawn
from the caller's brief and cross-checked against the architecture doc
`architecture-token-harness-load-soak.md` §6 (tooling) and §8 (P0 foundation
fixups), which mandate exactly these three code changes plus tests.

### PASS

- **[AC-1]** Correct stale `docs/performance.md` (lru_cache/alru_cache/maxsize=128/cache_info()/cache_clear() → real TTL-OrderedDict LRU stack) — rewritten at `docs/performance.md:10-197` and `docs/performance.md:514-541`. Every documented constant was verified against source:
  - 64-entry LRU → `DEFAULT_MAX_CACHE_ENTRIES: int = 64` (`core/jwks_cache.py:77`)
  - 32-stripe single-flight → `_DISCO_LOCK_STRIPES = 32` / `_JWKS_LOCK_STRIPES = 32` (`aio/token_validation.py:73-74`)
  - kid-miss cooldown 5s → `DEFAULT_KID_MISS_REFRESH_COOLDOWN_SECONDS: float = 5.0` (`core/jwks_cache.py:62`), env override `KID_MISS_REFRESH_COOLDOWN` (`core/jwks_cache.py:308`)
  - TTL defaults 3600s disco / 86400s JWKS → `core/jwks_cache.py:39-40`; env overrides `DISCO_CACHE_TTL`/`JWKS_CACHE_TTL` (`core/jwks_cache.py:153,162`)
  - clamp [60s, 86400s] → `MIN_CACHE_TTL_SECONDS=60.0` / `MAX_CACHE_TTL_SECONDS=86400.0` (`core/jwks_cache.py:41-42`)
  - per-loop `WeakKeyDictionary` locks → `aio/token_validation.py:12,75-86`
  - `OrderedDict` backing store + policy in `core/jwks_cache.py`, stacks in `sync/token_validation.py`/`aio/token_validation.py` → confirmed (`sync/token_validation.py:59`, `aio/token_validation.py:186`)
  - Breaking-change note "async clear_* now `async def`" → `aio/token_validation.py:165,315` are `async def`; sync counterparts are plain `def` (`sync/token_validation.py:123,248`)
  - Verified no residual `lru_cache`/`alru_cache` in the cache path (only unrelated `ssl_config.py:41`). Doc's honesty caveat ("sync stack not yet instrumented") matches reality: `grep get_cache_counters src/py_identity_model/sync/token_validation.py` → 0.
  Doc-only AC; no test tier applicable. Every factual claim traces to source (tier: n/a — documentation).

- **[AC-2]** Cache hit/miss/refresh counters on the async cache paths — `core/cache_metrics.py:1-116` (`CacheCounters`, `get_cache_counters`, singleton `CACHE_COUNTERS`), wired into `aio/token_validation.py`: disco hit (`:135`, fast-path + `:147` double-checked), disco miss (`:154`), jwks hit (`:230` + double-checked `:242`), jwks miss (`:249`), jwks refresh (`:288`, only on the branch that issues the upstream GET). Re-exported at top level (`__init__.py:247`, `__all__` entries). Tested at `src/tests/unit/test_cache_metrics.py:180` (miss→hit disco), `:200` (miss→hit jwks), `:219` (refresh counts upstream fetch), driving the real async cache functions with respx mocks (tier: integration — exercises the actual `_get_disco_response`/`_get_cached_jwks`/`_refresh_jwks` against mocked HTTP, asserting `route.call_count` proves hits skip the network). 15/15 pass locally.

- **[AC-3]** End-to-end `validate_token` warm-path micro-benchmark — `src/tests/benchmarks/test_benchmarks.py:214-311` (`test_bench_validate_token_warm_path`). Pre-warms disco+JWKS caches once under respx, then benches pure cache-hit `aio.validate_token` (RS256, full decode + claim validation), asserting `result["sub"] == "bench-user"`. Complements the pre-existing raw `pyjwt.decode` baseline (`:214`). Ran successfully (mean ~92us, 2665 rounds, group "validation") (tier: e2e micro-benchmark).

- **[AC-4]** Unit tests for the counters — `src/tests/unit/test_cache_metrics.py:40-140` (`TestCacheCountersUnit`): every `record_*` method, `snapshot` detachment (`:97`), `reset` (`:108`), thread-safety smoke with 8×1000 concurrent increments (`:124`), singleton identity (`:146`). Each test would fail if the corresponding counter/method were removed. 15/15 pass (tier: unit).

### FAIL

_(none)_

### PARTIAL

_(none)_

### SCOPE CREEP

- `src/tests/security/test_aio_cache_fetch_controls.py:1-393` (new, 393 lines) — a security-behavior pin-down suite for the three instrumented cache functions (scheme enforcement / SSRF, cooldown-sidecar eviction, refresh coalescing, double-checked-lock LRU recency). Not traceable to any AC; it exists because instrumenting these functions makes them mutation-security-gated. Non-blocking and defensible (protects the touched functions from mutation regressions), flagged for awareness.
- `tools/mutation_security_allowlist.txt:24-58` — 19 mutant allowlist entries for the instrumented functions (lock-stripe class-2 + log-only class-1 mutants). Adjunct to the security suite above; non-blocking, expected fallout of gating these functions.

### Architecture Violations

_(none)_ — Architecture doc §8 P0 lists exactly these three fixups ("correct `docs/performance.md`; add cache-hit/upstream-fetch instrumentation; add the e2e `validate_token` warm micro"); all three are delivered. §5 metrics surface ("JWKS + discovery cache-hit-rate (new counter)") and §2 "Cache instrumentation … hit/miss/refresh; aggregate across workers" are satisfied, including the per-process/aggregate-externally scoping the doc calls out (documented in `core/cache_metrics.py:9-14` and `docs/performance.md:191-194`).

### Summary
- Total ACs: 4

jamescrowley321:
### review-sentinel

Scope: T299 adds per-process cache observability counters (`core/cache_metrics.py`),
wires them into the three async cache-fetch functions in `aio/token_validation.py`
(`_get_disco_response`, `_get_cached_jwks`, `_refresh_jwks`), adds fail-closed tests,
and appends 20 waivers to `tools/mutation_security_allowlist.txt`. I verified the
SHIPPED/wired state and independently reproduced the mutation gate.

None.

### WARN (should fix)
None.

### INFO (acceptable risk)

- `src/py_identity_model/aio/token_validation.py:138,150,154,230,240,247,291` —
  The change is purely additive observability. Each `get_cache_counters().record_*()`
  is a standalone statement inserted alongside existing logic; it removes/alters no
  control flow, return value, policy, lock, or cooldown. The instrumented functions
  remain wired to the public entrypoint: `validate_token` (line 531) →
  `_discover_and_resolve_key` (line 578) → cached path calls all three
  (`_get_disco_response`, `_get_cached_jwks`, `_refresh_jwks`). The injected-`http_client`
  path deliberately bypasses the caches and is not counted, consistent with the docs.

- `src/py_identity_model/core/cache_metrics.py:1` — Counters store only integers under
  a dedicated leaf `threading.Lock`; `snapshot()` returns a detached copy. No URIs,
  tokens, kids, or other sensitive material is stored or exposed. The `logger.info`
  lines log only `jwks_uri` (a public, discovery-derived URL), not a secret. No
  credential-exposure vector.

- `tools/mutation_security_allowlist.txt:26-58` — All 20 T299 waivers verified as
  genuinely-equivalent survivors, NOT silenced controls. I generated the mutmut
  mutant source and read each waived mutant by exact name, then reran the mutation
  gate under the real test selection (`src/tests/security` + `src/tests/unit`):
    * **Stripe-selection (2):** `x__get_disco_response__mutmut_19` and
      `x__refresh_jwks__mutmut_3` replace the fetch-lock key with `None`
      (`_get_*_fetch_lock(None)`). `hash(None) % 32` still yields a valid fixed
      stripe, so single-flight is preserved (actually over-broad); only which of
      the 32 stripes is chosen changes — a performance property, not a correctness
      or security one. Empirically **survived** (no test asserts distinct-URI
      stripe distribution for disco or for `_refresh_jwks`). Legitimately unkillable.
    * **Log-text / logger-arg (18):** `x__refresh_jwks__mutmut_11-17` and `42-52`
      each mutate only a `logger.info(...)` message string or argument (None msg,
      None arg, dropped format arg, XX-wrap, case-fold) on the two info lines. None
      touches the `and`/`or` empty-keys guard, the `>=` coalescing guard, the
      counter, or the return. Confirmed by reading every mutant body.
  Cross-checked: the changed functions produce exactly 20 in-scope survivors and all
  20 are waived — 0 unwaived survivors, so the gate is green — with no over-waiving of
  a killed mutant.

- Calibration evidence (not a finding): the sibling stripe-selection mutant
  `x__get_cached_jwks__mutmut_13` is NOT waived and is empirically **killed** by
  `test_two_distinct_uris_fetch_in_parallel_async` (which drives `_get_cached_jwks`
  for two distinct URIs under an arg-keyed lock fixture and asserts parallel timing).
  The allowlist waives exactly the two unkillable stripe mutants and leaves the
  killable one to a real fail-closed test — a correctly calibrated allowlist, not a
  blanket waiver.

- `src/tests/security/test_aio_cache_fetch_controls.py` — Fail-closed tests exist for
  the actual controls in these functions: HTTPS scheme enforcement / policy
  thread-through (SSRF / downgrade), cooldown-sidecar eviction (unbounded-growth /
  memory leak keyed by attacker URIs), empty-keys retained-cache fallback
  (`and`->`or`), refresh coalescing guard (`>=`->`>`), and double-checked-lock /
  LRU recency. Deleting any of these controls breaks a test.

### Summary
- BLOCK: 0 | WARN: 0 | INFO: 6
- Overall: PASS

jamescrowley321:
### review-viper

## Review: Red Team (Viper)

### Attack Surface

- **`aio/token_validation.py` — async discovery/JWKS cache fetch path.** The diff adds
  observability-counter calls (`get_cache_counters().record_*()`) to the three module-level
  singleton cache functions used during token validation: `_get_disco_response`,
  `_get_cached_jwks`, `_refresh_jwks`. This is the code that resolves the signing key on every
  resource-server request, so any weakening of scheme enforcement, single-flight, cooldown, or
  the empty-keys fallback here would be directly exploitable.
- **`core/cache_metrics.py` (new).** Process-global `CacheCounters` singleton; pure integer
  tallies behind a leaf `threading.Lock`. Re-exported from the package root.
- **`tools/mutation_security_allowlist.txt` — mutation-gate waivers.** 20 new waived mutants for
  `aio.token_validation`. A wrongly-justified waiver could let a real auth-bypass /
  cache-poisoning regression slip past the fail-closed mutation gate undetected.
- Test/benchmark/doc changes (`test_aio_cache_fetch_controls.py`, `test_cache_metrics.py`,
  `test_benchmarks.py`, `docs/performance.md`) — no production runtime surface; the new security
  test suite *adds* fail-closed coverage.

### Findings

No exploitable findings.

Detailed adversarial verification performed:

1. **Counter instrumentation is side-effect-only.** Every `record_*` call is an O(1) integer
   increment; none appear in a conditional, return, or key-resolution expression. Hits are
   recorded on the two cache-hit branches (fast-path + double-checked), miss immediately before
   the upstream `get_discovery_document` / `get_jwks`, and refresh only on the branch that issues
   the upstream GET (after the coalescing early-return). No control flow into `find_key_by_kid`,
   `decode_with_config`, scheme enforcement, `>=` coalescing, `and` empty-keys fallback, or the
   kid-miss/signature-failure cooldown is altered. No auth-bypass, algorithm-confusion, or
   cache-poisoning path is introduced.

2. **No sensitive data leaves via the counters.** `snapshot()` returns only five integers
   (`disco_hits/misses`, `jwks_hits/misses/refreshes`) — no URIs, kids, issuers, or token
   material. Not an information-disclosure vector.

3. **New hot-path lock is not a DoS vector.** The counter's `threading.Lock` is held only for a
   single `+= 1` with no I/O; against the RSA/EC signature verify that dominates each request it
   is negligible. No constructible amplification.

4. **All 20 mutation-gate waivers verified benign by regenerating the mutants (mutmut 3.7.0)
   and diffing each against the original:**
   - `x__get_disco_response__mutmut_19` → `_get_disco_fetch_lock(None)` and
     `x__refresh_jwks__mutmut_3` → `_get_jwks_fetch_lock(None)`. Confirmed: only the *stripe
     selection* changes (`hash(None)%32` instead of `hash(key)%32`). Same-URI fetches still
     serialise; the primary amplification control (kid-miss / signature-failure cooldown gate)
     is upstream and untouched. Not security-observable.
   - `x__refresh_jwks__mutmut_11-17` and `42-52` → confirmed pure `logger.info(...)` message-text
     / logger-argument mutations (None-arg, dropped-arg, string-empty, case-flip) on the two
     info-log lines. Zero effect on control flow, return value, or counters.

5. **The waivers do NOT hide any security-relevant mutant.** The genuinely security-critical
   mutants for `_refresh_jwks` — `#9` (`>=`→`>` coalescing guard), `#18-25`
   (`policy=None` / `require_https=None` / dropped `policy` / `address=None` — SSRF & HTTPS
   downgrade), `#30` (`cooldown=None` sidecar leak), `#36/37/40/41` (`and`→`or` empty-keys
   fallback) — all fall *outside* the waived numeric ranges and remain gate-enforced, pinned by
   the new fail-closed suite in `src/tests/security/test_aio_cache_fetch_controls.py`
   (scheme enforcement, cooldown-sidecar eviction, empty-keys fallback, coalescing boundary,
   double-checked-lock + LRU recency).

6. **Live cache-security controls are unchanged by the diff.** The `policy=policy` /
   `require_https` thread-through, the `entry.cached_at >= request_time` single-flight guard, the
   `response.is_successful and not response.keys` retained-cache fallback, and the
   `_kid_miss_last_attempt` cooldown are byte-for-byte identical to pre-diff aside from the
   inserted counter calls. No regression in single-flight or cooldown behaviour.

### Summary
- Attack surface elements: 4
- Overall: PASS

Issue #464, opened 2026-08-01
test(harness): boot fastapi-identity-model as a real resource server (TH-1.2, T301)


jamescrowley321:
Part of #462 (E2E Token Harness). See planning `epics-token-harness.md`.

Pytest fixture that runs `packages/fastapi-identity-model` (TokenValidationMiddleware + require_scope) under uvicorn with protected endpoints, reachable over httpx. Reuses `conformance/app_fastapi.py`.

DoD: real-HTTP integration test.

Issue #465, opened 2026-08-01
test(harness): token correctness assertion matrix — pytest (TH-1.3, T302)


jamescrowley321:
Part of #462 (Token Validation Harness). See planning `epics-token-harness.md`.

pytest matrix: send one token per class (valid, expired, wrong-iss, wrong-aud, tampered-sig, unknown-kid, wrong-alg, DPoP-bound) from each provider; assert 200 for valid and 401 with the expected error per invalid class, nothing silently accepted. **Deterministic** — the load/soak swarm is a separate suite (TH-1.5 #474). Depends: TH-1.1, TH-1.2.

PR #522, opened 2026-08-13
test(harness): boot fastapi-identity-model as a real resource server (TH-1.2)

jamescrowley321:
## Summary
Greenfield minimal FastAPI resource server booted as a **real** out-of-process uvicorn server, reachable over httpx — the TH-1.2 RS test fixture.

- `src/tests/harness/rs_app.py` — greenfield `TokenValidationMiddleware(audience, require_access_token_marker) + require_scope`-guarded app (NOT the RP login router in `conformance/app_fastapi.py`). Config via `RS_*` env so uvicorn `--workers N` forked workers re-import an identical module-level `app`. Routes: `/health` (excluded), `/protected` (guarded by `require_scope`), `/whoami`.
- `src/tests/harness/rs_server.py` — fastapi-free `boot_rs(...)` context manager: free-port bind, `subprocess` uvicorn by import-string, `/health` readiness poll over httpx, teardown terminate/kill. Kept fastapi-free so pyrefly type-checks it.
- `src/tests/integration/test_rs_boot.py` — real-HTTP DoD: valid CC token → 200; missing/malformed auth → uniform 401; missing scope → 403; `/health` unauth → 200; `workers=2` → 200. `pytest.importorskip("fastapi_identity_model")`.
- `Makefile` — `test-harness-rs` target (node-oidc Docker + `uv run --all-packages`).
- `pyproject.toml` — pyrefly-exclude only `rs_app.py` (fastapi unresolved in lint env).

Refs #464
Epic #462

## Review Findings Addressed
Independent blind + edge + acceptance review (converged in 1 iteration):
- **P0 (1, fixed):** post-kill `proc.wait()` could raise `TimeoutExpired` out of the `finally` — wrapped in `contextlib.suppress`.
- **P1 (2, fixed):** uvicorn stdout PIPE never drained → ~64KB buffer deadlock → redirected to a `tempfile.TemporaryFile` (subsumes the earlier `stdout.close` ResourceWarning fix); empty `RS_EXCLUDED_PATHS` dropped the `/health` exclusion → empty/unset now falls back to the middleware default.
- **Acceptance:** 6/6 ACs PASS, mergeable. `/whoami` + F-07 marker path untested (belong to later TH-1.3 tasks).

## Test plan
- [x] Unit tests pass (`make lint`: ruff + format + pyrefly + pytest coverage)
- [x] Integration tests pass (`make test-harness-rs` → 6/6 real-HTTP against live node-oidc)
- [x] Lint passes
- [ ] CI passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,15 @@ test-integration-keycloak: ## Run integration tests against Keycloak
 		(docker compose -f test-fixtures/keycloak/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/keycloak/docker-compose.yml down
 
+.PHONY: test-harness-rs
+test-harness-rs: ## Boot the RS (uvicorn) against node-oidc and run the TH-1.2 real-HTTP proof
+	@echo "Starting node-oidc-provider fixture..."
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml up -d --build --wait
+	@echo "Booting fastapi-identity-model RS under uvicorn (real HTTP)..."
+	uv run --all-packages pytest src/tests/integration/test_rs_boot.py -m integration --env-file=.env.node-oidc -v || \
+		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,10 @@ project_excludes = [
     "examples/descope/",
     "packages/fastapi-identity-model/",
     "conformance/",
+    # Harness RS app imports the fastapi stack, unresolved in the root lint env
+    # (same rationale as packages/). Exercised via `make test-harness-rs`
+    # under `uv run --all-packages`.
+    "src/tests/harness/rs_app.py",
     # release tooling imports python-semantic-release, which the dev env
     # doesn't install; validated by the release pipeline itself
     "tools/",
--- a/src/tests/harness/__init__.py
+++ b/src/tests/harness/__init__.py
@@ -1,0 +1,5 @@
+"""Token-blaster harness helpers (TH-1.x).
+
+Test-only infrastructure for booting py-identity-model as a real resource
+server and driving it over HTTP. Not part of the shipped library.
+"""
--- a/src/tests/harness/rs_app.py
+++ b/src/tests/harness/rs_app.py
@@ -33,11 +33,16 @@ def _truthy(value: str | None) -> bool:
 
 def _excluded_paths() -> list[str] | None:
     """Parse ``RS_EXCLUDED_PATHS`` (comma-separated) or return ``None`` so the
-    middleware falls back to its own defaults (which include ``/health``)."""
+    middleware falls back to its own defaults (which include ``/health``).
+
+    An unset OR empty ``RS_EXCLUDED_PATHS`` both fall back to the middleware
+    defaults. An empty string (which is what ``boot_rs(excluded_paths=[])``
+    serializes to) must NOT be read as an explicit "exclude nothing" list —
+    that would drop the ``/health`` exclusion and break the readiness poll."""
     raw = os.environ.get("RS_EXCLUDED_PATHS")
-    if raw is None:
+    if not raw:
         return None
-    return [p for p in raw.split(",") if p]
+    return [p for p in raw.split(",") if p] or None
 
 
 def create_app() -> FastAPI:
--- a/src/tests/harness/rs_app.py
+++ b/src/tests/harness/rs_app.py
@@ -1,0 +1,90 @@
+"""Greenfield FastAPI resource server for the token-blaster harness (TH-1.2).
+
+A minimal RS that mounts ``TokenValidationMiddleware`` and guards routes with
+``require_scope``. All configuration comes from ``RS_*`` environment variables
+so uvicorn ``--workers N`` forked workers each re-import this module and rebuild
+an identical app (the workers are forked *after* import-string resolution, so
+they cannot share a closure — only the env).
+
+This is the ONLY harness module that imports FastAPI / fastapi-identity-model.
+It is excluded from the root pyrefly lint env (see ``[tool.pyrefly]
+project_excludes`` in ``pyproject.toml``) because that env installs only
+py-identity-model and cannot resolve the fastapi stack; it is exercised via
+``make test-harness-rs`` under ``uv run --all-packages``.
+"""
+
+from __future__ import annotations
+
+import os
+
+from fastapi import Depends, FastAPI
+
+from fastapi_identity_model import (
+    Claims,
+    CurrentUser,
+    TokenValidationMiddleware,
+    require_scope,
+)
+
+
+def _truthy(value: str | None) -> bool:
+    return (value or "").strip().lower() in {"1", "true", "yes", "on"}
+
+
+def _excluded_paths() -> list[str] | None:
+    """Parse ``RS_EXCLUDED_PATHS`` (comma-separated) or return ``None`` so the
+    middleware falls back to its own defaults (which include ``/health``)."""
+    raw = os.environ.get("RS_EXCLUDED_PATHS")
+    if raw is None:
+        return None
+    return [p for p in raw.split(",") if p]
+
+
+def create_app() -> FastAPI:
+    """Build the resource-server app from ``RS_*`` env config.
+
+    Required env: ``RS_DISCOVERY_URL``, ``RS_AUDIENCE``.
+    Optional env: ``RS_REQUIRE_SCOPE`` (default ``api``),
+    ``RS_REQUIRE_ACCESS_TOKEN_MARKER`` (bool, F-07 opt-in),
+    ``RS_EXCLUDED_PATHS`` (comma-separated).
+    """
+    discovery_url = os.environ["RS_DISCOVERY_URL"]
+    audience = os.environ["RS_AUDIENCE"]
+    required_scope = os.environ.get("RS_REQUIRE_SCOPE", "api")
+
+    app = FastAPI(title="token-harness-rs")
+    app.add_middleware(
+        TokenValidationMiddleware,
+        discovery_url=discovery_url,
+        audience=audience,
+        excluded_paths=_excluded_paths(),
+        require_access_token_marker=_truthy(
+            os.environ.get("RS_REQUIRE_ACCESS_TOKEN_MARKER")
+        ),
+    )
+
+    @app.get("/health")
+    def health() -> dict:
+        # Excluded from auth (middleware default) — the readiness probe target.
+        return {"status": "ok"}
+
+    @app.get("/protected", dependencies=[Depends(require_scope(required_scope))])
+    def protected(claims: Claims) -> dict:
+        # Reached only after the middleware validated the token AND the token
+        # carries ``required_scope`` (else require_scope raises 403).
+        return {
+            "sub": claims.get("sub"),
+            "scope": claims.get("scope") or claims.get("scp"),
+        }
+
+    @app.get("/whoami")
+    def whoami(user: CurrentUser) -> dict:
+        identity = user.identity
+        return {"name": identity.name if identity is not None else None}
+
+    return app
+
+
+# Module-level app so uvicorn can boot it by import string
+# (``src.tests.harness.rs_app:app``) and forked workers re-import identically.
+app = create_app()
--- a/src/tests/harness/rs_server.py
+++ b/src/tests/harness/rs_server.py
@@ -15,8 +15,9 @@ from pathlib import Path
 import socket
 import subprocess
 import sys
+import tempfile
 import time
-from typing import TYPE_CHECKING
+from typing import IO, TYPE_CHECKING
 
 import httpx
 
@@ -43,12 +44,23 @@ def _free_port() -> int:
         return int(sock.getsockname()[1])
 
 
-def _drain(proc: subprocess.Popen[str]) -> str:
-    return proc.stdout.read() if proc.stdout is not None else ""
+def _drain(log: IO[str]) -> str:
+    """Return everything uvicorn has written to its capture file so far.
+
+    Reading from a seekable temp file (not a live PIPE) never blocks: the child
+    writes to its own inherited fd and we snapshot from position 0. This also
+    means the child can emit an arbitrarily large traceback without filling an
+    OS pipe buffer and deadlocking on ``write()`` (blind/edge SHOULD-FIX).
+    """
+    try:
+        log.seek(0)
+        return log.read()
+    except (OSError, ValueError):  # pragma: no cover - defensive, file closed
+        return ""
 
 
 def _wait_for_health(
-    proc: subprocess.Popen[str], base_url: str, timeout: float
+    proc: subprocess.Popen[str], log: IO[str], base_url: str, timeout: float
 ) -> None:
     deadline = time.monotonic() + timeout
     last_error = "no attempt made"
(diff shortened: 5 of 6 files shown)


jamescrowley321:
### Review artifact: review-blind

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/tests/harness/rs_server.py:268`] The subprocess is spawned with `stdout=subprocess.PIPE` and `stderr=subprocess.STDOUT`, but the pipe is never read while the server is running — it is only drained via `_drain()` on failure/exit. If uvicorn (or the app) writes more than the OS pipe-buffer size (~64KB) during a healthy run, the child blocks on `write()` and can deadlock the server while tests keep hitting it. `--log-level warning` only reduces, not eliminates, this risk (a crashing worker can emit a large traceback).
- [`src/tests/harness/rs_app.py:89`] `_excluded_paths()` returns `None` only when `RS_EXCLUDED_PATHS` is *unset*. If it is set to an empty string, `"".split(",")` filtered yields `[]`, which is passed to the middleware as an explicit "exclude nothing" list — dropping the `/health` exclusion. The readiness poll then hits an authenticated `/health`, gets 401, and `boot_rs` times out. This directly contradicts the docstring's stated intent of falling back to defaults, and `boot_rs` produces exactly this empty string when called with `excluded_paths=[]` (`",".join([]) == ""`).

### NITPICK
- [`src/tests/harness/rs_server.py:218`] On the timeout branch `_wait_for_health` calls `proc.terminate()` and then raises; `boot_rs`'s `finally` immediately calls `proc.terminate()` again and waits. The double-terminate is harmless but redundant, and `_drain()` (`proc.stdout.read()`) in the raised message can block until EOF if the just-terminated process is slow to exit or ignores SIGTERM.
- [`src/tests/harness/rs_app.py:142`] `app = create_app()` runs at import time, so a missing `RS_DISCOVERY_URL`/`RS_AUDIENCE` surfaces as a bare `KeyError` inside the uvicorn worker rather than a clear "required env var" message; the failure only shows up indirectly as an unhealthy-boot timeout.
- [`src/tests/harness/rs_server.py:194`] `_drain` reads the full stdout with a blocking `proc.stdout.read()` and no timeout; combined with a child that does not close its stdout promptly, an error path could hang instead of surfacing the intended diagnostic.

jamescrowley321:
### Review artifact: review-edge-case

## Review: Edge Case Hunter

### Findings

| `src/tests/harness/rs_server.py:143` | uvicorn writes >~64KB to stdout/stderr while server is yielded (no reader drains the PIPE) | redirect stdout to a temp file: `stdout=tempfile.TemporaryFile(mode="w+")` instead of `PIPE` | [DEGRADED] OS pipe buffer fills, uvicorn blocks on write, requests hang/time out |
| `src/tests/harness/rs_server.py:150` | SIGKILL'd process not reaped within grace (uninterruptible I/O / stuck worker) | wrap in try: `except subprocess.TimeoutExpired: pass` | [CRASH] `TimeoutExpired` raised from `finally`, masks test result / errors teardown |
| `src/tests/harness/rs_server.py:194` `_drain` | `proc.stdout.read()` blocks until all forked `--workers` close the pipe after terminate | read with timeout or `proc.communicate(timeout=...)` | [DEGRADED] health-timeout diagnostic hangs if a worker ignores SIGTERM |

### Summary
- Unhandled paths found: 3

jamescrowley321:
### Review artifact: review-acceptance

## Review: Acceptance Auditor

**Spec:** #464 — `test(harness): boot fastapi-identity-model as a real resource server (TH-1.2, T301)`
**Enforcement doc:** `architecture-token-harness-load-soak.md` §2 (RS-under-test / TH-1.2), §1 + §7 (greenfield RS boot).

### Extracted acceptance criteria
- **AC-1** — A pytest harness boots `fastapi-identity-model` (`TokenValidationMiddleware` + `require_scope`) under **uvicorn** (issue body; arch §2).
- **AC-2** — The RS exposes **protected endpoints** guarded by `require_scope` (issue body; arch §2 "`require_scope`-guarded routes").
- **AC-3** — The running RS is **reachable/driven over httpx** (issue body).
- **AC-4** — **DoD: real-HTTP integration test** (not an in-process ASGI shim) (issue body).
- **AC-5** — Middleware is configured with `audience=…` and `require_access_token_marker=…` (arch §2, TH-1.2 row).
- **AC-6** — RS boots under **uvicorn `--workers N`** (arch §2 + §1).

### PASS
- [AC-1] RS harness boots the shipped middleware + `require_scope` under real uvicorn — app built in `src/tests/harness/rs_app.py:95` (`create_app`, `TokenValidationMiddleware` at :108, `require_scope` at :123); launched as a `python -m uvicorn` subprocess by import string in `src/tests/harness/rs_server.py:268`. Tested at `src/tests/integration/test_rs_boot.py:361` (and every other test boots via `boot_rs`) (tier: integration).
- [AC-2] Protected route `/protected` guarded by `Depends(require_scope(...))` at `rs_app.py:123`. Positive path tested at `test_rs_boot.py:361` (200 + claims echo); negative scope path (403) tested at `test_rs_boot.py:408`, asserting the real `require_scope` detail (`Required scope '…' not found`, verified in `dependencies.py:283`) (tier: integration).
- [AC-3] Every assertion drives the RS via `httpx.get(base_url/...)` (`test_rs_boot.py:374, 389, 398, 420, 432, 450`); readiness poll in `rs_server.py:210` also uses httpx (tier: integration).
- [AC-4] DoD proof is a real out-of-process uvicorn boot against a **live node-oidc issuer** (`test_rs_boot.py` module, `pytestmark = pytest.mark.integration` at :333; wired via `make test-harness-rs` in `Makefile:9`). Not a `TestClient`/ASGI shim. `integration` marker is registered (`pyproject.toml:246`) and `.env.node-oidc` + the node-oidc docker fixture both exist. Confirmed `uvicorn` (0.47.0) and `fastapi_identity_model` resolve under `uv run --all-packages` (tier: integration/e2e).
- [AC-5] Middleware constructed with `audience=os.environ["RS_AUDIENCE"]` and `require_access_token_marker=…` (`rs_app.py:108-116`); both are real `TokenValidationMiddleware.__init__` params (`middleware.py:101,103`). `audience` is fully exercised (all boots pass `RS_AUDIENCE="urn:test:api"`; missing-auth/malformed 401 uniformity asserted at `test_rs_boot.py:386,395`). Note: the marker is wired and constructed in every boot but only in its default-off state; the marker=True (F-07 ID-token-as-access rejection) path is out of scope for #464 and belongs to the TH-1.3 correctness matrix (#465) — not a gap for this issue (tier: integration).
- [AC-6] Multi-worker boot supported via `workers` kwarg → `--workers N` (`rs_server.py:243,277-279`); exercised with `workers=2` and a 200 assertion at `test_rs_boot.py:438` (tier: integration).

### FAIL
- (none)

### PARTIAL
- (none)

### SCOPE CREEP
- `src/tests/harness/rs_app.py:132` — the `/whoami` route (using `CurrentUser`) is not traceable to any AC and has **no test** covering it. Non-blocking; it demonstrates the `CurrentUser` identity surface but is dead relative to the DoD. Consider removing or adding coverage.

### Architecture Violations
- (none) — Note on an *intentional, doc-sanctioned* divergence from the literal issue text: the issue body says "Reuses `conformance/app_fastapi.py`", but architecture §1 and §7 explicitly **override** this ("the RS boot itself is greenfield — `app_fastapi.py` is the RP *login* router, not `TokenValidationMiddleware + require_scope`"). The diff correctly implements a **greenfield** RS (`rs_app.py`) per the authoritative architecture doc. Following the issue's literal wording would have been wrong; this is compliant.
- Enforcement note (compliant): the pyrefly exclusion of `rs_app.py` (`pyproject.toml:32`) is justified (fastapi stack unresolved in root lint env, same rationale as `packages/`) and `rs_server.py` is deliberately kept free of fastapi imports so it stays inside the lint env — matches the module docstrings. `filterwarnings=error` (`pyproject.toml`) makes the explicit `proc.stdout.close()` in `rs_server.py:303` load-bearing, and it is present.

### Summary
- Total ACs: 6
- Pass: 6 | Fail: 0 | Partial: 0
- Scope creep: 1 (non-blocking, untested `/whoami`)
- Verdict: **MERGEABLE** — all acceptance criteria implemented and covered by real-HTTP integration tests.

jamescrowley321:
### Review artifact: review-blind-delta

Re-review of the FIX diff (`/tmp/pim-T301/.claude/review-diff.patch`) against the
current source of `rs_server.py` and `rs_app.py`. Scope: verify the two addressed
findings and hunt for NEW regressions introduced by the fix diff only.

### MUST FIX
None.

### SHOULD FIX
None.

Both prior findings are correctly resolved and I found no regression that warrants a fix:

- **Finding 1 (pipe-buffer deadlock -> tempfile).** `subprocess.Popen` given a real
  file object for `stdout` uses that object's `fileno()` for the child fd; the child
  writes raw bytes straight to the kernel fd. `text=True` only governs pipes that
  Popen *creates* itself, so with a file object passed in it is a harmless no-op and
  `proc.stdout` is `None` (which is exactly why the old `proc.stdout.close()` was
  correctly removed). The parent never writes through the `log` TextIOWrapper, so there
  is no dirty buffer to flush; `_drain` does `seek(0)` (lseek to offset 0) then `read()`,
  which decodes everything the child wrote. Parent reads what the child wrote — correct.
  The `with tempfile.TemporaryFile(...) as log:` block encloses the `Popen`, the `yield`,
  AND the whole teardown `finally`, so the capture file outlives every `_drain` call and
  is only closed/unlinked after the process is reaped. Lifetime is correct; no fd leak
  (`TemporaryFile` unlinks immediately on Linux, and the child's dup'd fd is released
  when the killed process exits).

- **Finding 2 (empty `RS_EXCLUDED_PATHS` -> defaults).** `if not raw: return None`
  now folds both unset and empty-string to `None`, and the trailing `... or None`
  also folds an all-separator string (e.g. `",,"` -> `[]`) to `None`. Since
  `boot_rs(excluded_paths=[])` serializes to `""`, it now correctly falls back to the
  middleware defaults that include `/health`, so the readiness poll keeps working.
  This means an explicit "exclude nothing" list is no longer expressible — but
  `boot_rs`'s docstring mandates keeping `/health`, so that is the intended and
  acceptable behavior, not a regression. A caller that passes a non-empty list omitting
  `/health` (e.g. `["/foo"]`) still breaks the poll, but that was true before the diff
  and is explicitly the caller's contract per the docstring.

Additional regression checks (all clean):
- If `Popen(...)` itself raises, `proc` is never bound but there is no proc to leak, and
  the `with tempfile` still closes the file — same shape as the pre-fix code.
- `_wait_for_health` calls `proc.terminate()` internally before raising; the outer
  `finally` then terminates+waits again (idempotent) so the process is still reaped.
- The new double-`wait` with `contextlib.suppress(subprocess.TimeoutExpired)` prevents a
  post-`kill()` timeout from raising out of `finally` and masking the real test result —
  correct, and `contextlib` is already imported.
- `stderr=subprocess.STDOUT` with `stdout=log` routes stderr to the same capture fd —
  fine.

### NITPICK
- `_drain` reads the capture file in the parent's text mode (locale-default encoding),
  while the child writes bytes in *its own* encoding. Under a C/ASCII locale a non-ASCII
  byte in a uvicorn traceback would raise `UnicodeDecodeError` — but that is a subclass
  of `ValueError`, which `_drain`'s `except (OSError, ValueError)` already swallows, so
  the worst case is an empty diagnostic string rather than a crash. Optional: pass
  `errors="replace"`/`encoding="utf-8"` to `TemporaryFile` to preserve the traceback.
  Non-blocking.

### Verdict
The fixes are correct and introduce no new correctness, security, or resource
regressions. Approve.

jamescrowley321:
### Review artifact: review-edge-case-delta

# Review: Edge Case Hunter (delta)

Scope: the FIX diff `/tmp/pim-T301/.claude/review-diff.patch` against current
`src/tests/harness/rs_server.py` and `src/tests/harness/rs_app.py`. Verifying
the three addressed findings and hunting for regressions introduced by this diff
only.

## Verdict: none (no CRASH / no DEGRADED regressions)

The three prior findings are correctly fixed and the fix introduces no new
crash or degradation path. Details below.

---

## Verification of the three fixes

### 1. [was P0 CRASH] teardown `proc.wait()` after kill — FIXED, correct
The second `proc.wait(timeout=_TERMINATE_GRACE)` is now wrapped in
`contextlib.suppress(subprocess.TimeoutExpired)` (rs_server.py:173-174).
`contextlib` is already imported (line 12). Walking the whole `finally` block
for any residual raise:
- `proc.terminate()` — CPython `send_signal` is a no-op once `returncode` is set
  and swallows `ProcessLookupError`; does not raise here.
- first `proc.wait(...)` — only `TimeoutExpired`, which is caught.
- `proc.kill()` — same no-op/`ProcessLookupError`-safe path as terminate.
- second `proc.wait(...)` — `TimeoutExpired` now suppressed; nothing else it can
  raise in this context.
Teardown can no longer propagate out of `finally` and mask the real test result.
Correct.

### 2. [was P1 DEGRADED] PIPE buffer deadlock — FIXED, correct
Child stdout/stderr now redirect to `tempfile.TemporaryFile(mode="w+")`
(rs_server.py:138-159). Confirmed the parent reads child-written bytes correctly:
- Popen passes the file's `fileno()` to the child via dup2; the child writes to a
  **regular file**, which never blocks on `write()` — the ~64KB pipe backpressure
  is structurally gone, including the multi-worker case (all `--workers N`
  children share one inherited open-file-description / offset; regular-file writes
  don't deadlock).
- `_drain()` does `log.seek(0); log.read()` (lines 55-59). The `seek(0)`
  invalidates the parent `TextIOWrapper`'s decode buffer and repositions the
  underlying binary layer, so each drain re-reads fresh from the OS to current
  EOF. The parent never *writes* to `log`, so there is no unflushed parent-side
  buffer to worry about. Reads return the child's bytes correctly.
- `_drain` guards `OSError`/`ValueError` (closed-file), so a drain after teardown
  can't crash.

### 3. [was P1] empty `RS_EXCLUDED_PATHS` dropped `/health` — FIXED, correct
`_excluded_paths()` now returns `None` for both unset and empty (`if not raw`)
and additionally collapses an all-empty list to `None`
(`[...] or None`, rs_app.py:43-45). So `boot_rs(excluded_paths=[])`
(serializes to `""`) → `None` → middleware default (keeps `/health`), and the
readiness poll still passes. An explicit non-empty list is preserved unchanged.
Correct, and no behavior change for legitimate non-empty lists.

---

## Regression / edge sweep of the fix diff

Checked and cleared:
- **`text=True` + a text-mode file object as `stdout`** — not a bug. Popen's
  `text` only governs streams Popen itself creates (PIPE); when an actual file
  object is passed, Popen uses its `fileno()` and `text` is a no-op. The child
  writes locale/UTF-8 text bytes to the fd and the `mode="w+"` parent decodes
  them on read. Works. (`text=True` is now vestigial here — a harmless nit, not a
  defect.)
- **Resource / fd leak** — `with tempfile.TemporaryFile(...)` encloses the whole
  try/finally, so the file is closed *and* unlinked after teardown; no lingering
  fd, no `ResourceWarning` that `filterwarnings=error` could promote. If `Popen`
  itself raises, `proc` is never referenced by a finally and the `with` still
  closes the file — no leak.
- **Concurrent read during shutdown** — on the timeout branch `_drain` may read
  while the child is still writing (post-`terminate`, pre-exit); reading a regular
  file concurrently with writes yields a possibly-partial snapshot but never
  blocks or errors. Benign, not a regression.
- **Generator/`GeneratorExit`** — on early consumer teardown, contextmanager
  throws into the `yield`; the `finally` runs teardown and the `with tempfile`
  then closes cleanly.
- **Signature threading** — `_drain(log)` / `_wait_for_health(proc, log, ...)`
  are updated consistently at both call sites (lines 71, 85, 161); `IO` is
  imported (line 20).

## Residual (non-blocking) notes
- After `proc.kill()` + suppressed second timeout, a SIGKILL'd-but-unreaped child
  is left orphaned while the temp file is unlinked; the child keeps the fd until
  it dies. This is the accepted tradeoff of the CRASH fix (never raise from
  teardown) and is fine for a test harness.
- `text=True` is now dead/vestigial on this Popen call. Cosmetic only.

No changes required.

PR #521, opened 2026-08-13
test(harness): unified multi-provider TokenSource minter + mock OP + forged corpus (TH-1.1)

jamescrowley321:
## Summary
Introduces the shared token-harness package `src/tests/harness/` (TH-1.1), the foundation for the E2E Token-Blaster harness (epic #462):

- **`token_source.py`** — one unified `TokenSource.mint(provider, grant, tenant, scopes, malform)` consolidating the per-provider conftest minters (client-credentials, JWT/opaque access, auth-code) and the Descope multi-tenant access-key exchange minter. Capability/credential-gated via `provider_matrix.detect_capabilities`, raising typed `HarnessCapabilityError`/`HarnessCredentialError` that conftest translates to `pytest.skip`. Includes the pre-minted **replay pool** (`prime_pool`, xdist FileLock + JSON cache) tracking `expires_at` for the 300s TTL.
- **`mock_op.py`** — a framework-free ASGI OIDC stub holding a known RS256+ES256 signing key. Serves discovery/JWKS/token/optional introspection, with T311 failure-injection tunables (latency, 429/Retry-After, 5xx, rotate-key-on-command, no-store, empty/oversized JWKS) via a mutable `MockOPControls` + loopback-only `/_control` route.
- **`corpus.py`** — the forged/negative token corpus off the mock-OP key: tampered-sig, unknown-kid, wrong-alg/`alg:none`, wrong-iss/aud, id-as-access, cnf-bound-as-bearer, oversized/many-claim, multi-aud.
- **conftest** — additive session `token_source` + `harness_provider` fixtures; the 32 existing integration tests stay green (no minter fixtures rewritten).

Real Descope multi-tenant exchange (`/v1/auth/accesskey/exchange`) mirrors the identity-stack e2e helpers, credential-gated on `DESCOPE_PROJECT_ID`/`MANAGEMENT_KEY`/`BASE_URL`/`TEST_TENANT_ID`.

Refs #463
Epic #462

## Review Findings Addressed
3 adversarial reviewers (Blind + Edge + Acceptance). All P0 fixed:
- mock OP introspect non-numeric `exp` → try/except (`active:false`)
- mock OP `/_control` untyped setattr → `_set_control` type-checks vs `MockOPControls` fields
- `_peek_jwt_header` non-object JWT header → `isinstance(dict)` guard
- AC-3 Descope multi-tenant `dct`/`tenants` → real access-key exchange implemented
[...]

Issue #478, opened 2026-08-01
feat(dpop): RS-side DPoP proof verification + cnf.jkt enforcement (RFC 9449 §5–7)


jamescrowley321:
Part of epic #476 (FAPI 2.0 Security Profile — hardening & conformance). Sibling of #215 (mTLS cert-bound tokens). Builds on the client-side DPoP work in #94/#255.

## Problem

PIM ships DPoP **proof creation** (`core/dpop.py`) but has **no resource-server verifier**, so a DPoP-bound access token is accepted exactly like a plain bearer token. Confirmed in code (2026-08-01 review, see `conformance/adding-conformance-profiles.md`, PR #477):

- `core/dpop.py` is client-side only — builds proofs, computes `ath`, no verifier.
- `core/token_validation_logic.decode_with_config` → `decode_and_validate_jwt` validates key/alg/aud/iss/sub/exp only; **never reads `cnf`**.
- `fastapi-identity-model` middleware (`_extract_bearer_token`) only accepts the `Bearer` scheme and does **zero** binding check.

Result: an RS built on PIM does not honor the sender-constraint. A stolen bound token, or a token presented without a matching proof, is accepted. This is the core FAPI-2 RS guarantee and it is the load-bearing gap behind the E2E token harness (#462–#474).

**Scoping note:** #215 is scoped to *mTLS* (`x5t#S256`) cert-bound tokens. This issue is the **DPoP (`jkt`)** counterpart — deliberately split so each can land and be tested independently. The FAPI-2 *client* conformance plan (#475) only tests token *acquisition* (client-side DPoP, already shipped), so #475 does **not** block on this — but the E2E harness does.

## Scope / design

- New `core/dpop_verify.py` — server counterpart to `core/dpop.py`. Given the request (method + URL), the `DPoP` proof header, and the access token, verify per RFC 9449 §4.3/§5–7:
  - header `typ == "dpop+jwt"`, `alg` is asymmetric + allow-listed (reuse `_SUPPORTED_ALGORITHMS`);
  - signature validates against the embedded `jwk`;
  - `htm` == request method, `htu` == request URI (normalized: no query/fragment, per §4.3);
  - `iat` within an acceptance window; `jti` unseen within a replay window (pluggable replay cache, default in-memory);
  - if an access token is presented, `ath` == base64url(SHA-256(token));
[...]

Issue #474, opened 2026-08-01
test(harness): load/soak suite — Locust (TH-1.5, T311)


jamescrowley321:
Part of #462 (Token Validation Harness). See planning `epics-token-harness.md`.

Locust swarms the `fastapi-identity-model` resource server with a **pre-minted token pool** (mint once via TH-1.1, replay many — real IdPs rate-limit per-request minting; JWT validation is stateless so replay is valid), spanning issuers/tenants. Reports RPS / latency percentiles / error-rate + JWKS-discovery cache hit-rate; short CI profile + longer nightly soak (feeds #271).

Split of responsibility: **pytest owns correctness + deterministic behavior proofs (LRU/mix-up/private_key_jwt); Locust owns load/soak.** Depends: TH-1.1, TH-1.2.

Issue #466, opened 2026-08-01
test(harness): make test-harness target + CI gating (TH-1.4, T303)


jamescrowley321:
Part of #462 (Token Validation Harness). See planning `epics-token-harness.md`.

`make test-harness` runs the pytest correctness matrix (TH-1.3) **and** the Locust CI load profile (TH-1.5 #474); node-oidc+keycloak always, Ory/Descope when secrets present; documented nightly hook. Related: #271. Depends: TH-1.3, TH-1.5.

PR #523, opened 2026-08-13
test(harness): token correctness assertion matrix through booted RS (TH-1.3)

jamescrowley321:
## Summary

Adds the **token correctness assertion matrix** (TH-1.3, #465): send one token per class through the **booted resource server** (T301 `boot_rs`) and assert 200 for valid, **401 with the uniform F-18 body** per invalid class, and that **nothing is silently accepted**.

- `src/tests/harness/mock_op_server.py` (NEW, pyrefly-excluded) — `serve_mock_op()` context manager serves the framework-free `MockOP` over **real localhost HTTP** (in-process uvicorn daemon thread, `interface="asgi3"`, `lifespan="off"`) so the out-of-process booted RS can fetch discovery + JWKS, while the forged corpus is minted off the same in-process signing key.
- `src/tests/integration/test_correctness_matrix.py` (NEW):
  - **Group A — mock-OP forged-corpus matrix (self-contained, no Docker):** valid→200; the 8 validation-failure classes (expired, nbf-future, wrong-iss, wrong-aud, tampered-sig, unknown-kid, wrong-alg, alg:none) → **uniform 401** (F-18); `id_as_access` → 401 via the always-on negative defense (F-07) and 403/401 via the opt-in marker gate; `cnf_bound` → **200** (F-02 accepted-today contract, tracked by #478); `nothing_silently_accepted` guard.
  - **Group B — real-issuer node-oidc leg (real-HTTP DoD, gated to node-oidc like `test_rs_boot`):** real client-credentials token → 200; wrong-audience RS → 401 generic. Proves the contract against a **real IdP over real HTTP** through the booted RS.
- `Makefile` — `test-harness-matrix` target (node-oidc Docker + `uv run --all-packages … --env-file=.env.node-oidc`).

Refs #465
Epic #462

> **Stacked PR.** This branch builds on the T300 minter (#521) and T301 RS-boot (#522) foundations, neither yet merged to `main`. The diff therefore includes those files until #521 and #522 merge first; once they do, this PR collapses to the T302-only changes (`mock_op_server.py`, `test_correctness_matrix.py`, the `test-harness-matrix` target). No production `core/` code is touched.

## Review Findings Addressed

Three fresh-context review agents (blind + edge-case + acceptance): **0 MUST-FIX / 0 FAIL**. Applied SHOULD-FIX items across two review-fix rounds (`fe012df`, `5fced0d`):
- `serve_mock_op`: `_ServerThread` captures uvicorn startup exceptions; `_wait_for_discovery` fast-fails on thread death instead of polling the full 30s (blind+edge convergent).
- Bounded shutdown with a leaked-thread / port-wedge guard, gated on `sys.exc_info()[0] is None` so a failing test body keeps its own exception as the headline (delta-review regression fix).
- `_granted_scopes`: return `[]` for non-str / non-sequence scope claims (edge CRASH guard).
- Docstring/import nitpicks.

Acceptance: 5 PASS / 1 PARTIAL (AC-2 "from each provider" satisfied via the mock-OP full matrix + node-oidc valid/wrong-aud; Keycloak/Ory/Descope real legs gated per the design's forge-off-mock-OP strategy).

## Test plan
- [x] Unit tests pass (`make lint` — ruff/format/pyrefly/pytest 80% cov, green after every commit)
- [x] Integration tests pass — `make test-harness-matrix`: **20 passed** (Group A forged-corpus matrix + Group B real node-oidc leg through the booted RS over real HTTP), `MAKE_EXIT=0`
- [x] Lint passes
- [x] Independent review agents passed (0 MUST-FIX / 0 FAIL)
- [ ] CI passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,15 @@ test-harness-rs: ## Boot the RS (uvicorn) against node-oidc and run the TH-1.2 r
 		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
 
+.PHONY: test-harness-matrix
+test-harness-matrix: ## Run the TH-1.3 token correctness matrix (mock-OP forged corpus + node-oidc leg) through the booted RS
+	@echo "Starting node-oidc-provider fixture..."
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml up -d --build --wait
+	@echo "Running the correctness matrix through the booted RS (real HTTP)..."
+	uv run --all-packages pytest src/tests/integration/test_correctness_matrix.py -m integration --env-file=.env.node-oidc -v || \
+		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,15 @@ test-integration-keycloak: ## Run integration tests against Keycloak
 		(docker compose -f test-fixtures/keycloak/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/keycloak/docker-compose.yml down
 
+.PHONY: test-harness-rs
+test-harness-rs: ## Boot the RS (uvicorn) against node-oidc and run the TH-1.2 real-HTTP proof
+	@echo "Starting node-oidc-provider fixture..."
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml up -d --build --wait
+	@echo "Booting fastapi-identity-model RS under uvicorn (real HTTP)..."
+	uv run --all-packages pytest src/tests/integration/test_rs_boot.py -m integration --env-file=.env.node-oidc -v || \
+		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,10 @@ project_excludes = [
     "examples/descope/",
     "packages/fastapi-identity-model/",
     "conformance/",
+    # Harness RS app imports the fastapi stack, unresolved in the root lint env
+    # (same rationale as packages/). Exercised via `make test-harness-rs`
+    # under `uv run --all-packages`.
+    "src/tests/harness/rs_app.py",
     # release tooling imports python-semantic-release, which the dev env
     # doesn't install; validated by the release pipeline itself
     "tools/",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,10 @@ ignore = [
 
 [tool.ruff.lint.per-file-ignores]
 "src/tests/**/*.py" = ["S101"]   # 1858 assert violations — all test assertions, 0 in production code
+"src/tests/harness/**/*.py" = [
+    "S101",    # pytest assertions
+    "PLR0913", # minter / mock-OP APIs are intentionally parameter-rich (mint, sign, mint_access_token)
+]
 "examples/**/*.py" = ["S101"]    # 73 assert violations in example integration tests
 "packages/**/tests/**/*.py" = [
     "S101",    # pytest assertions
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,6 +120,9 @@ project_excludes = [
     # (same rationale as packages/). Exercised via `make test-harness-rs`
     # under `uv run --all-packages`.
     "src/tests/harness/rs_app.py",
+    # Serves the mock OP under uvicorn (deferred import), unresolved in the root
+    # lint env; exercised via `make test-harness-matrix` under `--all-packages`.
+    "src/tests/harness/mock_op_server.py",
     # release tooling imports python-semantic-release, which the dev env
     # doesn't install; validated by the release pipeline itself
     "tools/",
--- a/src/tests/harness/__init__.py
+++ b/src/tests/harness/__init__.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 from .corpus import CORPUS_AUDIENCE, ForgedToken, build_corpus
 from .mock_op import MockOP, MockOPControls, SigningKey
+from .mock_op_server import serve_mock_op
 from .token_source import (
     Grant,
     HarnessCapabilityError,
@@ -48,4 +49,5 @@ __all__ = [
     "TokenSource",
     "build_corpus",
     "prime_pool",
+    "serve_mock_op",
 ]
--- a/src/tests/harness/__init__.py
+++ b/src/tests/harness/__init__.py
@@ -1,0 +1,51 @@
+"""Token-blaster harness shared infrastructure (epic #462, TH-1.1..TH-1.5).
+
+A single home for the pieces T300/T301/T311 share:
+
+* :class:`TokenSource` — the unified multi-provider minter (``mint``) plus the
+  pre-minted :class:`ReplayPool`.
+* :class:`MockOP` — a controllable, framework-free ASGI OpenID Provider holding
+  a known signing key (valid tokens, forged corpus, T311 failure injection).
+* :func:`build_corpus` — the forged / negative token corpus.
+"""
+
+from __future__ import annotations
+
+from .corpus import CORPUS_AUDIENCE, ForgedToken, build_corpus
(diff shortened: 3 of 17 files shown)


jamescrowley321:
## Review artifact: blind

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/tests/harness/mock_op_server.py:125-140`] TOCTOU race on port selection: `_free_port()` opens a socket, reads the assigned port, and closes it, then uvicorn re-binds that same port a few lines later. Between the close and the re-bind another process (or a parallel `pytest-xdist` worker running the same helper) can grab the port. The bind failure then surfaces only as an opaque "did not serve discovery" timeout after the full 30s, not as a clear "address in use" error.
- [`src/tests/harness/mock_op_server.py:141-148`] A uvicorn bind/startup failure inside the daemon thread is silently swallowed. `server.run()` raises inside the thread, the thread dies, and `_wait_for_discovery` polls until the 30s deadline before raising a generic `RuntimeError`. The real error (bind refused, port in use, ASGI misconfig) is lost, making every startup failure look identical and slow to diagnose.
- [`src/tests/harness/mock_op_server.py:148`] `thread.join(timeout)` return value is not checked. If the server never honors `should_exit` (e.g. it was still mid-startup when the flag was set), `join` returns silently after `timeout` seconds and the daemon thread — still bound to the loopback port — leaks past the `with` block, which can wedge the next `serve_mock_op` that reuses the same port range.

### NITPICK
- [`src/tests/harness/mock_op_server.py:102`] The `timeout` parameter is overloaded for two unrelated waits: the discovery readiness poll and the thread-join shutdown. A slow startup and a slow shutdown are tuned by the same knob, so you cannot bound them independently.
- [`src/tests/integration/test_correctness_matrix.py:209-211`] Import ordering is inconsistent: `from collections import namedtuple` precedes `import time`, mixing `from`-imports and plain `import`s within the stdlib group; ruff/isort would normally group plain imports first.
- [`src/tests/integration/test_correctness_matrix.py:276`] Fixture docstring says "single uvicorn subprocess," but `serve_mock_op` explicitly runs the mock OP in an in-process daemon thread (per `mock_op_server.py`'s own docstring). The comment contradicts the mechanism it describes.

jamescrowley321:
## Review artifact: edge-case

## Review: Edge Case Hunter

### Findings

| `src/tests/harness/mock_op_server.py:151` | uvicorn thread dies at startup (e.g. port TOCTOU collision, bind error) | `if not thread.is_alive(): raise RuntimeError(...)` (pass thread into poll loop) | [DEGRADED] Polls the full `timeout` (30s) before raising though server is already dead; sibling `_wait_for_health` fails fast via `proc.poll()` |
| `src/tests/harness/mock_op_server.py:126` | caller passes `controls` with `discovery_status != 200` or high `latency_seconds` | gate readiness on `/jwks` or skip poll when discovery failure is injected | [DEGRADED] `_wait_for_discovery` never sees a 200, so `serve_mock_op` hangs the full timeout then raises despite server being healthy |
| `src/tests/integration/test_correctness_matrix.py:434` | `scp`/`scope` claim is a non-string, non-iterable value (e.g. JSON int) | `if not isinstance(raw, (str, list)): return []` | [CRASH] `for s in raw` raises `TypeError` when the real IdP mints a non-list scope claim |

### Summary
- Unhandled paths found: 3

jamescrowley321:
## Review artifact: acceptance

## Review: Acceptance Auditor

**Spec (#465 · TH-1.3):** "pytest matrix: send one token per class (valid, expired, wrong-iss, wrong-aud, tampered-sig, unknown-kid, wrong-alg, DPoP-bound) from each provider; assert 200 for valid and 401 with the expected error per invalid class, nothing silently accepted. **Deterministic** — the load/soak swarm is a separate suite (TH-1.5 #474). Depends: TH-1.1, TH-1.2."

Extracted ACs:
- **AC-1** — Send one token per class through the RS; classes = valid, expired, wrong-iss, wrong-aud, tampered-sig, unknown-kid, wrong-alg, DPoP-bound.
- **AC-2** — From each provider.
- **AC-3** — Assert 200 for valid.
- **AC-4** — Assert 401 with the expected error per invalid class.
- **AC-5** — Nothing silently accepted.
- **AC-6** — Deterministic (not the load/soak swarm).

### PASS
- [AC-1] Every required class is present in the forged corpus and driven through the **booted** RS (real uvicorn subprocess + real HTTP): valid, expired, wrong-iss, wrong-aud, tampered-sig, unknown-kid, wrong-alg, DPoP-bound(=`cnf_bound`). Corpus built at `src/tests/harness/corpus.py:82` (build_corpus), keyed to the served `MockOP`. Exercised at `src/tests/integration/test_correctness_matrix.py:96` (accepted classes), `:118` (validation failures parametrized over all 8 reject-classes), `:151` (nothing-silently-accepted iterates the whole corpus). Extras beyond the 8 named classes (nbf_future, alg_none, id_as_access, oversized, multi_aud_untrusted) are all traceable to design §3/§4-S8. **(tier: integration)** — real localhost mock OP over HTTP (`src/tests/harness/mock_op_server.py:32` serve_mock_op) + booted RS (`src/tests/harness/rs_server.py:98` boot_rs); node-oidc leg is Docker+HTTP. Note: DPoP-bound (`cnf_bound`) is asserted as **200 accepted-today** (`test_correctness_matrix.py:133`), which is the *opposite* direction from the terse spec's "401 per invalid class." This is a deliberate, documented deviation grounded in the architecture doc's **F-02** resolution ("cnf-bound-as-bearer accepted today; assert & flag", design §0/§4-S8/§9-#478). The test flags it and references #478 for the future flip. Following the authoritative design doc over the terse spec is the correct call — PASS.
- [AC-3] Valid token → 200 asserted three ways: `test_accepted_classes_reach_protected` (`test_correctness_matrix.py:94`), `test_valid_token_body_echoes_claims` (`:100`, also verifies sub/scope echo), and real-issuer `test_real_issuer_valid_token_accepted` (`:265`). Middleware/route path confirmed at `packages/fastapi-identity-model/fastapi_identity_model/middleware.py` + `src/tests/harness/rs_app.py:76` (require_scope-guarded `/protected`). **(tier: integration)**
- [AC-4] Each invalid class → 401 with the expected error. All 8 validation-failure classes collapse to the uniform F-18 body `{"detail": "Invalid or unauthorized token"}` — asserted at `test_correctness_matrix.py:118` (per-class) and `:127` (`test_f18_body_uniform_across_all_failure_stages`). Body matches the middleware's real `_GENERIC_401_DETAIL` (`middleware.py:57`, `:150` `_unauthorized`). The token-type-confusion class (`id_as_access`) → its distinct 401 detail `"ID token cannot be used as an access token"` (`test_correctness_matrix.py:137`), matching `middleware.py:185`. Real-issuer negative `test_real_issuer_wrong_aud_rejected` (`:283`) proves the uniform 401 against a live IdP. "Expected error per class" is correctly the F-18 uniform body (per design §0/§4-S8). **(tier: integration)**
- [AC-5] `test_nothing_silently_accepted` (`test_correctness_matrix.py:151`) iterates the *entire* corpus and asserts every class outside the explicit accepted set returns non-200, with a failure message naming the leaked class. **(tier: integration)**
- [AC-6] Deterministic: single token per class, no load generator, fixed assertions, `pytestmark = pytest.mark.integration`. Load/soak is explicitly left to TH-1.5 (module docstring `test_correctness_matrix.py:1`). **(tier: integration)**

### FAIL
- (none)

### PARTIAL
- [AC-2] "from each provider" — The controllable **mock OP** (full forged matrix, Group A) and **node-oidc** (Group B: valid + wrong-aud, `test_correctness_matrix.py:265`/`:283`) are covered. The other providers named in the design's issuer set — **Keycloak** (local, per design §2), **Ory**, **Descope** (CI-only) — are **not exercised** by this matrix: the `node_oidc_discovery_url` fixture (`:245`) `pytest.skip`s every non-node-oidc provider (`_is_node_oidc_fixture`, `:225`), and even Keycloak (also local) is skipped by that gate. Design §3 justifies forging the negative matrix off the single mock OP (real OPs won't emit invalid tokens) and scopes remote providers to CI, so this is a defensible engineering decision — but the literal "from each provider" real-issuer coverage is met for exactly one real IdP. The design's own "reuse committed expired tokens + cross-provider kid-mismatch" real-negative path (§3) is also not wired into this matrix. Not blocking on its own merits given the design's forging strategy, but the multi-real-provider leg is incomplete versus the literal AC.

### SCOPE CREEP
- (none) — `mock_op_server.py` (new) is required infrastructure: the out-of-process booted RS fetches discovery/JWKS over real HTTP and cannot reach an in-process ASGI mock, so the mock OP must be served on a real loopback port keyed to the *same* signing instance the corpus is built from (`mock_op_server.py:1` docstring). Extra corpus classes (F-07/F-18/F-02 cases) trace to design §4-S8. The `boot_rs` `noqa: PLR0913` removal (`rs_server.py:90`) is an incidental lint cleanup, harmless.

### Architecture Violations
- (none mandatory) Minor observation: design §6.3 tags TH-1.3 as a "**required PR gate**," but the diff wires it only as the `make test-harness-matrix` target (`Makefile:74`) gated behind Docker + `--all-packages` + `importorskip`; no CI job wiring appears in this diff. CI gating is plausibly TH-1.4 (#466) scope. `mock_op_server.py` added to `[tool.pyrefly] project_excludes` (`pyproject.toml:123`) follows the existing deferred-import pattern (mirrors `rs_app.py`) — not a violation.

### Summary
- Total ACs: 6
- Pass: 5 | Fail: 0 | Partial: 1

Issue #271, opened 2026-03-29
Story 11.5: Nightly Provider Validation CI


jamescrowley321:
## Epic 11: Cloud Provider Integration Testing — Story 11.5

**As a** library maintainer,
**I want** a nightly CI workflow that runs integration tests against real cloud providers,
**So that** provider drift is detected automatically and cassettes stay current.

**FR covered:** FR-PIM-27

### Scope

1. GitHub Actions nightly workflow (`cron: '0 6 * * *'` + `workflow_dispatch`)
2. Separate jobs per provider (Cognito, Entra, Auth0) for isolation
3. Provider secrets stored in GitHub Actions secrets
4. On failure: auto-create GH issue labeled `provider-drift` with provider name, test, error
5. Transient failure handling: retry once after 5 min, label "[Transient]"
6. Manual cassette refresh workflow: `workflow_dispatch` with `record-provider` input → re-record cassettes, open PR

### Acceptance Criteria

- [ ] Nightly workflow runs on schedule and via manual dispatch
- [ ] Each provider runs in separate job
- [ ] Secrets configured for all 3 providers
- [ ] Test failure creates labeled GH issue with details
- [ ] Network errors trigger retry before issue creation
- [ ] Manual cassette refresh workflow records and opens PR
- [ ] All-pass run completes silently (no issues created)

### Dependencies
- Stories 11.2–11.4 (provider tests exist)
- Provider secrets configured in GitHub Actions

jamescrowley321:
Marking deferred 2026-05-21. Current workstream focus narrowed to conformance certification (#331) and the security/testing tail (#380, #381, #398, #399, #403). This epic's cassette + cloud-provider integration work remains valid but is not in the active queue. Will revisit once the conformance milestone closes.

jamescrowley321:
Superseded. Multi-provider integration was implemented differently — a live `provider_matrix.py` against Ory, Keycloak, node-oidc-provider and Descope (see `src/tests/integration/`), not the cassette-based Cognito/Entra/Auth0 approach this story proposed. Closing as superseded; re-file against the current live-matrix infra if this provider's coverage is still wanted.

PR #524, opened 2026-08-13
test(load): Locust load/soak suite S1-S12 vs booted RS + mock OP (TH-1.5)

jamescrowley321:
## Summary
Adds the TH-1.5 Locust load/soak harness (design §4/§5) driving the booted resource server (T301) with the pre-minted replay pool (T300) and the controllable mock OP (T300) for failure injection.

- **Merged the T299 cache-observability foundation** (`core/cache_metrics.py` `get_cache_counters().snapshot()`, per-process hit/miss/refresh counters in `aio/token_validation.py`) — the cache-hit-rate surface TH-1.5 reports.
- **Counter accuracy:** miss/refresh now count *attempted* upstream fetches including failed (429/5xx) round-trips, gated on a side-effect-free scheme re-check (`_upstream_fetch_attempted`) rather than `response.is_successful`, so an upstream outage no longer reads as zero upstream volume.
- **Mock-OP upstream-fetch counters** (per-path discovery/jwks/token/introspect) + read/reset `/_stats` route; **RS `/metrics`** unauthenticated readout of the cache counters.
- **`src/tests/load/`**: `scenarios.py` (S1–S12 table + profile tiers), `pool.py` (replay pool + post-expiry-401 = expected classification), `locustfile.py` (out-of-process Locust — parent never imports locust to avoid gevent/asyncio deadlock), `runner.py`, `test_load_ci_short.py` (real-Locust `@integration` DoD).
- Profiles: **CI_SHORT** = S1,S2,S3,S6,S8; **DIAGNOSTIC** = S5/S9/S10; **NIGHTLY** = S4/S7/S11/S12 (S4 needs a >60s TTL rollover — the cache has a hard 60s floor).
- `pyproject.toml` `load` dependency-group (locust, not a runtime/optional dep), pyrefly excludes for locust-importing modules, ruff per-file S101/PLR2004; `Makefile` `test-harness-load` (+ nightly stub).

## Stacking
Branched on the T300+T301+T302 harness stack (#521/#522/#523) and merges the T299 counters branch (#520); all four are still open/unmerged, so the diff vs `main` includes those foundation files until the owner merges the stack. Review T311's own contribution as `src/tests/load/**` + the mock_op/rs_app counter readouts + the `_upstream_fetch_attempted` counter-accuracy fix.

Refs #474, epic #462

## Review Findings Addressed
5 reviewers (Blind/Edge/Acceptance/Sentinel/Viper): **P0 = 0** (Sentinel 0 BLOCK, Acceptance 0 FAIL, Viper 0 findings — locust/gevent are a PEP 735 dev group, never shipped to pip consumers). 5 P1s fixed in `51e3561`+`6315c79`: counter under-count on 429/5xx, `cache_hit_rate` 1.0-when-empty → 0.0, S2 alg-cost ratio now computable (per-class p50/p95/p99 + p999), RS `/metrics` reuses fastapi `_default_excluded_paths()`, and honest CI_SHORT profile (S4→nightly, S10 scaffold flagged, stale "S1-S8" docstrings corrected). Delta re-review clean.

## Test plan
- [x] Unit tests pass (`make lint` — ruff + format + pyrefly + 80% coverage GREEN)
- [x] Integration tests pass — **real Locust DoD** `make test-harness-load`: 10 passed, S3 cold stampede coalesced to exactly 1 discovery + 1 JWKS fetch (single-flight), S1 warm all-hits, zero 5xx across all scenarios
- [x] Security gate passes — `make security-gate BASE=origin/main`: 158 in-scope mutants in the changed `_refresh_jwks`/counter functions all killed or waived-equivalent (4 log-string mutants waived with justification)
- [x] Lint passes
- [ ] CI passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -368,9 +368,36 @@ jobs:
           BASE_REF: ${{ github.base_ref || 'main' }}
         run: BASE="origin/$BASE_REF" make security-gate
 
+  load-smoke:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+        with:
+          python-version: 3.13
+
+      - name: Cache uv dependencies
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+        with:
+          path: ~/.cache/uv
+          key: uv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
+          restore-keys: |
+            uv-${{ runner.os }}-
+
+      - name: Install dependencies
+        run: make ci-setup
+
+      - name: Load smoke (CI-short profile — real Locust vs the booted RS)
+        run: make test-harness-load
+
   ci-complete:
     if: always()
-    needs: [lint, unit-tests, fastapi-tests, integration-tests-ory, integration-tests-descope, integration-tests-node-oidc, integration-tests-keycloak, example-tests, build, security-gate]
+    needs: [lint, unit-tests, fastapi-tests, integration-tests-ory, integration-tests-descope, integration-tests-node-oidc, integration-tests-keycloak, example-tests, build, security-gate, load-smoke]
     runs-on: ubuntu-latest
     steps:
       - name: Check CI results
@@ -386,6 +413,7 @@ jobs:
             "${{ needs.example-tests.result }}" \
             "${{ needs.build.result }}" \
             "${{ needs.security-gate.result }}" \
+            "${{ needs.load-smoke.result }}" \
           )
           for result in "${results[@]}"; do
             if [[ "$result" == "failure" || "$result" == "cancelled" ]]; then
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,59 @@
+name: Nightly
+
+# Scheduled coverage that does not belong on the per-PR path:
+#  * full-ci   — re-runs the whole reusable CI matrix (all integration providers)
+#                on a schedule, so upstream provider / dependency drift is caught
+#                even on days with no commits.
+#  * load-soak — the long-window load/soak profile (S4 TTL-rollover + S7/S11/S12
+#                LRU-thrash / RSS-FD soak) that is too slow for the PR gate.
+
+on:
+  schedule:
+    - cron: '0 7 * * *'   # 07:00 UTC daily
+  workflow_dispatch:       # allow manual runs
+
+permissions:
+  contents: read
+
+concurrency:
+  group: nightly-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  full-ci:
+    name: Full CI (all integrations)
+    uses: ./.github/workflows/ci.yml
+    with:
+      run-descope-integration: true
+      run-example-tests: true
+      run-build: true
+      run-ory-integration: true
+    secrets: inherit
+
+  load-soak:
+    name: Load soak (NIGHTLY profile)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
+        with:
+          python-version: 3.13
+
+      - name: Cache uv dependencies
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+        with:
+          path: ~/.cache/uv
+          key: uv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
+          restore-keys: |
+            uv-${{ runner.os }}-
+
+      - name: Install dependencies
+        run: make ci-setup
+
+      - name: Load soak (S4/S7/S11/S12 — real Locust vs the booted RS)
+        run: make test-harness-load-nightly
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -134,6 +134,9 @@ make test-integration-local # Requires .env.local file
 # Run example integration tests (spins up Docker containers)
 make test-examples
 
+# Run the TH-1.5 CI-short load profile (real Locust vs the booted RS + mock OP)
+make test-harness-load       # Opt-in `load` group (locust); Locust runs out-of-process
+
 # Run all tests including examples
 make test-all
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,8 +100,9 @@ src/py_identity_model/
    - Core functions are pure (no I/O) and return result objects or raise exceptions
 
 3. **Caching Strategy**
-   - Discovery documents: Cached per-process with `functools.lru_cache` (sync) and `async_lru.alru_cache` (async)
-   - JWKS keys: Cached to avoid repeated fetches during token validation
+   - Discovery documents and JWKS are cached per-process with an explicit **TTL + LRU** cache (an `OrderedDict` bounded to `DEFAULT_MAX_CACHE_ENTRIES` = 64), **not** `functools.lru_cache`/`async_lru.alru_cache`. Cache *policy* lives in `core/jwks_cache.py`; the sync/async cache stacks live in `sync/token_validation.py` and `aio/token_validation.py`.
+   - Entries expire on a TTL resolved from `Cache-Control: max-age` → env override (`DISCO_CACHE_TTL` / `JWKS_CACHE_TTL`) → built-in default, clamped to `[60s, 86400s]`. Concurrent misses are coalesced into a single upstream fetch via per-URI striped locks.
+   - Per-process cache hit/miss/refresh counters are exposed via `core/cache_metrics.py` (`get_cache_counters()`, re-exported at the top level); the async cache paths are instrumented. Clear caches with `clear_discovery_cache()` / `clear_jwks_cache()` (async variants must be `await`ed).
    - All caches are per-process, not shared across processes
 
 4. **Error Handling**
--- a/Makefile
+++ b/Makefile
@@ -87,12 +87,10 @@ test-harness-load: ## Run the TH-1.5 CI-short load profile (real Locust vs the b
 		-m integration -p no:benchmark -v
 
 .PHONY: test-harness-load-nightly
-test-harness-load-nightly: ## (nightly, NOT CI-gated) Long LRU-thrash / RSS-FD soak profile (S7/S11/S12)
-	@echo "Nightly soak profile (design §4 S7/S11/S12) — a documented, not-scheduled"
-	@echo "hook (#271). Run on demand once the SLO baseline is calibrated:"
-	@echo "  uv run --group load --all-packages python -c \\"
-	@echo "    'from src.tests.load.runner import run_profile; from src.tests.load.scenarios import Profile;\\"
-	@echo "     [print(r) for r in run_profile(Profile.NIGHTLY)]'"
+test-harness-load-nightly: ## (nightly) Long TTL-rollover / LRU-thrash / RSS-FD soak profile (S4/S7/S11/S12)
+	@echo "Driving the NIGHTLY soak profile (design §4 S4/S7/S11/S12) through the booted RS..."
+	uv run --group load --all-packages pytest src/tests/load/test_load_nightly.py \
+		-m integration -p no:benchmark -v
 
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,15 @@ test-harness-rs: ## Boot the RS (uvicorn) against node-oidc and run the TH-1.2 r
 		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
 
+.PHONY: test-harness-matrix
+test-harness-matrix: ## Run the TH-1.3 token correctness matrix (mock-OP forged corpus + node-oidc leg) through the booted RS
+	@echo "Starting node-oidc-provider fixture..."
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml up -d --build --wait
+	@echo "Running the correctness matrix through the booted RS (real HTTP)..."
+	uv run --all-packages pytest src/tests/integration/test_correctness_matrix.py -m integration --env-file=.env.node-oidc -v || \
+		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,20 @@ test-harness-matrix: ## Run the TH-1.3 token correctness matrix (mock-OP forged 
 		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
 
+.PHONY: test-harness-load
+test-harness-load: ## Run the TH-1.5 CI-short load profile (real Locust vs the booted RS + mock OP)
+	@echo "Driving the CI-short Locust profile through the booted RS (real HTTP)..."
+	uv run --group load --all-packages pytest src/tests/load/test_load_ci_short.py \
+		-m integration -p no:benchmark -v
+
+.PHONY: test-harness-load-nightly
+test-harness-load-nightly: ## (nightly, NOT CI-gated) Long LRU-thrash / RSS-FD soak profile (S7/S11/S12)
+	@echo "Nightly soak profile (design §4 S7/S11/S12) — a documented, not-scheduled"
+	@echo "hook (#271). Run on demand once the SLO baseline is calibrated:"
+	@echo "  uv run --group load --all-packages python -c \\"
+	@echo "    'from src.tests.load.runner import run_profile; from src.tests.load.scenarios import Profile;\\"
+	@echo "     [print(r) for r in run_profile(Profile.NIGHTLY)]'"
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,15 @@ test-integration-keycloak: ## Run integration tests against Keycloak
 		(docker compose -f test-fixtures/keycloak/docker-compose.yml down && exit 1)
 	docker compose -f test-fixtures/keycloak/docker-compose.yml down
 
+.PHONY: test-harness-rs
+test-harness-rs: ## Boot the RS (uvicorn) against node-oidc and run the TH-1.2 real-HTTP proof
+	@echo "Starting node-oidc-provider fixture..."
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml up -d --build --wait
+	@echo "Booting fastapi-identity-model RS under uvicorn (real HTTP)..."
+	uv run --all-packages pytest src/tests/integration/test_rs_boot.py -m integration --env-file=.env.node-oidc -v || \
+		(docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down && exit 1)
+	docker compose -f test-fixtures/node-oidc-provider/docker-compose.yml down
+
 .PHONY: test-benchmark
 test-benchmark: ## Run benchmarks
 	uv run pytest src/tests/benchmarks -v --benchmark-only --benchmark-sort=name
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -10,112 +10,108 @@ py-identity-model is designed for high performance with built-in caching and sup
 
 ### Discovery Document Caching
 
-Discovery documents are cached automatically to avoid repeated HTTP requests:
+Discovery documents and JWKS are cached in-process to keep the steady-state
+request path off the network. The cache is **not** `functools.lru_cache` /
+`async_lru.alru_cache` — those were replaced (v3.0.0) with an explicit
+**TTL + LRU** cache so entries expire on a schedule, key rotation is handled
+automatically, and concurrent misses are coalesced into a single upstream
+fetch. The cache *policy* lives in `core/jwks_cache.py`; the sync and async
+cache stacks live in `sync/token_validation.py` and `aio/token_validation.py`.
 
-**Sync Implementation:**
-```python
-from functools import lru_cache
+### What is cached
 
-@lru_cache(maxsize=128)
-def _get_disco_response(disco_doc_address: str) -> DiscoveryDocumentResponse:
-    """Cached discovery document fetching"""
-    request = DiscoveryDocumentRequest(address=disco_doc_address)
-    return get_discovery_document(request)
-```
+| Cache | Cache key | Backing store | Default TTL | Env override |
+|-------|-----------|---------------|-------------|--------------|
+| Discovery document | `(address, require_https)` | `OrderedDict` | 3600s (1 hour) | `DISCO_CACHE_TTL` |
+| JWKS | `jwks_uri` | `OrderedDict` | 86400s (24 hours) | `JWKS_CACHE_TTL` |
 
-**Async Implementation:**
-```python
-from async_lru import alru_cache
+Both caches share the same policy machinery:
 
-@alru_cache(maxsize=128)
-async def _get_disco_response(disco_doc_address: str) -> DiscoveryDocumentResponse:
-    """Cached async discovery document fetching"""
-    request = DiscoveryDocumentRequest(address=disco_doc_address)
-    return await get_discovery_document(request)
-```
+- **LRU eviction, 64 entries** (`DEFAULT_MAX_CACHE_ENTRIES`). The store is an
+  `OrderedDict`, not a plain dict, so a *read hit* refreshes recency
+  (`touch_cache_entry`) and eviction targets the least-recently-**used** entry.
+- **TTL resolution order.** For each entry the TTL is the first available of:
+  (1) the response's `Cache-Control: max-age`, (2) the environment override,
+  (3) the built-in default. The resolved value is clamped to
+  **[60s, 86400s]** (`MIN_CACHE_TTL_SECONDS` … `MAX_CACHE_TTL_SECONDS`); a bad
+  or out-of-range env value is clamped and logged, never crashes the request.
+- **Single-flight fetch.** Per-URI striped locks (32 stripes) ensure a cache
+  miss issues exactly one upstream fetch; concurrent requests for the same URI
+  wait on the stripe and then read the freshly-cached entry (a
+  *double-checked* hit) rather than each hitting the network.
+- **Kid-miss cooldown.** If a JWT's `kid` is absent from the cached JWKS the
+  cache is treated as stale and a refresh is forced (OP key rotation). This is
+  rate-limited per-URI — default 5s (`DEFAULT_KID_MISS_REFRESH_COOLDOWN_SECONDS`,
+  override `KID_MISS_REFRESH_COOLDOWN`) — so an attacker forging tokens with
+  random unknown kids cannot amplify inbound traffic into upstream JWKS fetches.
+- **Per-event-loop locks (async).** The async write/fetch locks are keyed on
+  the running event loop via `WeakKeyDictionary` (#399), so a test runner or
+  embed that creates a new loop per scope does not hit
+  `RuntimeError: <Lock> is bound to a different event loop`.
 
-### JWKS Caching
-
-JSON Web Key Sets are also cached:
-
-- **Cache Size**: 128 entries (both sync and async)
-- **Cache Key**: JWKS URL (from discovery document)
-- **Cache Duration**: Lifetime of the process
-- **Thread Safety**: Both implementations are thread-safe
-
-### Cache Behavior Differences
-
-| Aspect | Synchronous | Asynchronous |
-|--------|-------------|--------------|
-| Library | `functools.lru_cache` | `async_lru.alru_cache` |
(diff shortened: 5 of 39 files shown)


jamescrowley321:
### Review agent: review-blind

## Review: Blind Hunter

### MUST FIX
No findings.

### SHOULD FIX
- [`src/tests/harness/rs_app.py:_DEFAULT_EXCLUDED_PATHS`] The exclusion list `["/docs", "/openapi.json", "/health"]` is a hand-copied mirror of `fastapi_identity_model.config._default_excluded_paths`. The new code replaces the middleware's own default-resolution with this local copy whenever `_excluded_paths()` is empty. If the real middleware default ever gains/renames a path, this mirror silently drifts — a path the middleware intends to leave unauthenticated would start requiring a token (or vice-versa) with no test catching it. Import/reuse the real default instead of duplicating it.
- [`src/py_identity_model/aio/token_validation.py` — disco/jwks/refresh counting] Miss/refresh counters increment only when `response.is_successful`. An upstream `429`/`5xx` performs a real network round-trip but is recorded as neither a hit nor a miss, so the "upstream fetch volume" the counter exists to surface is undercounted precisely during upstream failure — the exact condition an operator watches this metric for. This also contradicts the shipped docs: `docs/performance.md` defines a miss as "request that had to fetch from upstream," while `cache_metrics.py` defines it as "fetched … and succeeded." Pick one definition; if failed fetches are intentionally excluded, the performance.md wording is wrong.
- [`src/tests/load/scenarios.py` — Scenario `S4`] Titled "TTL refresh under load (short TTL)" with a note about "cache TTL rollover mid-load," but it has no `setup` hook that shortens the TTL, so it is behaviorally identical to `S1` (plain warm valid run) and proves nothing about TTL rollover. Its `steady_state=True` gate therefore passes trivially. Same defect in `S10` ("blocking claims_validator") which installs no blocking validator. These scenarios give false coverage confidence.

### NITPICK
- [`src/tests/load/runner.py` — `_scrape_cache_metrics` / `cache_hit_rate`] `_scrape_cache_metrics` swallows all errors and returns `{}`; `cache_hit_rate` then returns `1.0` when totals are zero. A failed `/metrics` scrape thus reports a perfect 100% hit rate and would silently pass `test_s1_warm_cache_is_all_hits`' `>= 0.99` assertion — a broken scrape masquerades as a passing SLO.
- [`src/tests/load/runner.py` — `_run_locust`] `subprocess.run(..., timeout=...)` can raise `TimeoutExpired`, which is unhandled and gives no run context/cleanup; acceptable for a test harness but worth a guard.
- [`src/tests/harness/rs_app.py:metrics`] Adds an unauthenticated `/metrics` route exposing cache counters and explicitly forces it into the auth-exclusion set even when `RS_EXCLUDED_PATHS` overrides the defaults. Fine for a test-only harness app, but the "must never require a token regardless of the override" behavior is a deliberate auth bypass that should stay confined to the harness.

jamescrowley321:
### Review agent: review-edge-case

## Review: Edge Case Hunter

### Findings

| `src/py_identity_model/aio/token_validation.py:163` | Upstream disco/JWKS fetch does real network I/O but returns non-2xx (mock-OP 429/5xx injection, OP outage) | `if response.is_successful or response_did_network_io(response): record_disco_miss()` | [DEGRADED] Real failed fetch not counted as a miss; upstream-fetch volume undercounted |
| `src/tests/load/runner.py:1292` | Traffic hits an erroring upstream: 0 hits, 0 (uncounted) misses → `total == 0` | `return hits / total if total else 0.0` | [WRONG] `cache_hit_rate` reports 1.0 (100%) during a total upstream outage window |

### Summary
- Unhandled paths found: 2

### Notes

The `if response.is_successful:` guards added at
`aio/token_validation.py:163` (disco miss), `:260` (jwks miss) and `:308`
(jwks refresh) were intended to exclude *scheme-rejected* requests (which do
zero network I/O and return `is_successful=False` before any fetch). However
`is_successful` is a coarse proxy: a request that performed a real upstream
round trip but got a `429`/`5xx` (which the harness itself injects via
`MockOPControls.discovery_status`/`jwks_status`) also returns
`is_successful=False`, so it is silently dropped from the miss/refresh tally.
The counter therefore under-reports upstream-fetch volume exactly during the
failure-injection scenarios (S5/S6) it exists to observe, and — combined with
`LoadResult.cache_hit_rate` returning `1.0` when `total == 0`
(`runner.py:1292`) — makes an all-error window read as a perfect 100% cache
hit rate. No currently-active SLO gate consumes `cache_hit_rate`
(`GATES["warm"]`/`GATES["cold"]` are permissive), so this is a reported-metric
accuracy defect, not a test failure. Both verdicts are PLAUSIBLE.

All other changed source is defensive: `core/cache_metrics.py` counters are
lock-guarded O(1) int increments that cannot raise; the `record_*` calls read
`is_successful` (a plain bool field, always present); the pool/scenarios/mock-OP
harness code is internally consistent (every `EXPECTED_STATUS` class resolves to
a corpus or synthetic token, no `KeyError`). No crash-level unhandled paths.

jamescrowley321:
### Review agent: review-acceptance

## Review: Acceptance Auditor

Scope note: ACs derived from issue #474 body + the load/soak design doc §4 (scenarios) and §5 (metrics/SLOs), which the issue explicitly grounds. Per the task framing, the **current deliverable is the CI-short profile (S1, S2, S3, S4, S6, S8)**; scenarios S5/S9/S10 (DIAGNOSTIC), S7/S11/S12 (NIGHTLY), the node-oidc real-issuer replay leg, and the docs/README SLO baseline table are DEFERRED by design and are recorded as PARTIAL (deferral noted), not FAIL.

### PASS
- [AC-1] Locust replays a pre-minted token pool ("mint once, replay many") against the booted RS — implemented at `src/tests/load/pool.py:151` (`build_load_pool`) + `src/tests/load/locustfile.py:83` (`ReplayUser.hit_protected`), driven at `src/tests/load/runner.py:1314` (`_run_locust` real subprocess); tested at `src/tests/load/test_load_ci_short.py:59` (`test_scenario_drove_load_and_met_gates`, `num_requests > 0`) (tier: integration/e2e).
- [AC-3] Reports RPS — `locustfile.py:91` serializes `total_rps`; surfaced on `LoadResult.rps` (`runner.py:1265`); asserted indirectly via gates (tier: integration).
- [AC-4] Reports latency percentiles (p50/p95/p99) — `locustfile.py:92-94`; `LoadResult.p50/p95/p99_ms`; evaluated in `evaluate_gates` (`runner.py:1441`) (tier: integration).
- [AC-5] Reports error-rate by class (expected 401/403 excluded) — scoring at `locustfile.py:88-95` (fail only on status divergence), `LoadResult.error_rate`/`failures_by_class` (`runner.py:1281`); `EXPECTED_STATUS`/`EXPECTED_REJECTION_CLASSES` at `scenarios.py:1549` (tier: integration).
- [AC-6] Reports JWKS + discovery cache-hit-rate — RS `/metrics` endpoint at `src/tests/harness/rs_app.py:93`, scraped by `runner.py:1295` (`_scrape_cache_metrics`), `LoadResult.cache_hit_rate` (`runner.py:1286`); tested at `test_load_ci_short.py:88` (`test_s1_warm_cache_is_all_hits`, `>= 0.99`) (tier: integration).
- [AC-7] Short CI profile — `Profile.CI_SHORT` at `scenarios.py:107`, wired via `make test-harness-load` (`Makefile:26`); tested at `test_load_ci_short.py:52` (`test_every_ci_short_scenario_ran`) (tier: integration).
- [AC-9] Correctness-under-load split (Locust asserts no 500s + status-per-class) — `evaluate_gates` (`runner.py:1420`) enforces zero server-errors and zero steady-state divergence; tested at `test_load_ci_short.py:66` + `:59` (tier: integration).
- [AC-11] S1 warm-path RPS ceiling (valid RS256, hot cache) — `scenarios.py:1672`; warm-up + upstream-counter reset at `runner.py:1382`; tested at `test_load_ci_short.py:88` (zero upstream fetches in the measured window) (tier: integration).
- [AC-13] S3 cold-stampede single-flight (exactly 1 discovery + 1 JWKS fetch) — `scenarios.py:1688`; ground-truth via mock-OP `RequestStats` (`mock_op.py:736`) scraped into `upstream_stats`; tested at `test_load_ci_short.py:75` (`test_s3_cold_stampede_is_single_flight`, `== 1` each) and `test_mock_op.py:2642` (tier: integration).
- [AC-16] S6 kid-rotation storm + unknown-kid flood — `scenarios.py:1719` with active injection `_rotate_unpublished` (`scenarios.py:1655`, `op.rotate_keys(publish=False)`); no-500 asserted for S6 via `test_no_server_errors_anywhere`; cooldown-cap correctness is pytest-owned at `src/tests/unit/test_kid_miss_cooldown.py` (hardened here with `_pin_cooldown_high`) (tier: integration + unit).
- [AC-18] S8 rejection correctness & uniformity under contention — `scenarios.py:1740` (`MIXED_CLASSES`, `steady_state=True`); F-07 (`id_as_access`→401 via always-on negative check `middleware.py:184`), F-02 (`cnf_bound`→200 accepted-today), and cross-talk (`valid` must stay 200) all encoded in `EXPECTED_STATUS` and enforced by the steady-state gate; tested at `test_load_ci_short.py:59`+`:66` (tier: integration). Note: F-18 *uniform-body* equality is not asserted at load (status-only) — pytest-owned per the split.
- [AC-23] Cache-hit/upstream-fetch instrumentation counters (design §2/§5 "new counter") — `src/py_identity_model/core/cache_metrics.py` (new), async paths instrumented at `aio/token_validation.py:135/246/300`, exported at `__init__.py:328`; tested exhaustively at `src/tests/unit/test_cache_metrics.py` (unit + async-integration) and fail-closed security behaviour at `src/tests/security/test_aio_cache_fetch_controls.py` (tier: unit + integration).
- [AC-25] Zero 500s hard invariant — `evaluate_gates` (`runner.py:1429`) + tested at `test_load_ci_short.py:66` (`test_no_server_errors_anywhere`) and `locustfile.py:98` (`_count_server_errors`) (tier: integration).
- [Foundation §8-P0] `docs/performance.md` corrected to the real TTL+LRU cache stack and an e2e `validate_token` warm-path micro added — `docs/performance.md` rewrite + `src/tests/benchmarks/test_benchmarks.py:653` (`test_bench_validate_token_warm_path`) (tier: benchmark).

### PARTIAL
- [AC-2] Pool "spans issuers/tenants" — each scenario runs a **single** mock-OP issuer (`runner.py:1487` `_scenario_stack`), single-tenant. Multi-issuer/tenant (S12) and the real-provider replay are deferred to NIGHTLY / the node-oidc leg. Delivered: single-issuer forged+valid blend.
- [AC-8] Longer nightly soak (feeds #271) — `Profile.NIGHTLY` (S7/S11/S12) exists as data only; **no test drives it** and `make test-harness-load-nightly` (`Makefile:31`) is a documented echo-only no-op. Intentionally deferred; the soak itself (RSS/FD trend, multi-worker amplification) is unimplemented.
- [AC-10] Depends TH-1.1 (TokenSource) / TH-1.2 (RS boot) reuse — TH-1.2 fully reused (`boot_rs`/`rs_app`); the pool mints off the **mock OP corpus** (`build_corpus`) rather than the real-provider `TokenSource`. Real-IdP (node-oidc/Keycloak/Ory/Descope) minted-token replay is deferred.
- [AC-12] S2 algorithm cost matrix "report the RS256/ES256 ratio" — `scenarios.py:1680` drives both `valid` + `valid_es256`, but the run summary's `by_class` only serializes request/failure **counts**, not per-class latency (`locustfile.py:82-90`), so the cost ratio cannot be computed from the result. (The docstring at `locustfile.py:14` claims per-class latency percentiles are captured — they are not.) The alg *cost ratio* deliverable is not reported.
- [AC-14] S4 TTL-expiry refresh under load — `scenarios.py:1698` is in the CI-short deliverable set but applies **no short-TTL injection** and crosses no TTL boundary (defaults 3600s/86400s vs a 3s run), so the single-flight-refresh / no-error-spike behaviour it names is never exercised; the scenario self-documents TTL rollover as "calibrated in the test phase" (`scenarios.py:165`). Only generic no-500/steady-state is asserted.
- [AC-15] S5 provider-slowness head-of-line blocking — DIAGNOSTIC profile (`scenarios.py:1708`); injection hook present (`_inject_latency`) but not run/asserted by CI-short. Deferred by design.
- [AC-17] S7 connection-pool pressure + LRU thrash >64 issuers — NIGHTLY (`scenarios.py:1729`); single-issuer placeholder, no >64-issuer fan implemented. Deferred by design.
- [AC-19] S9 discovery `no-store` pathology — DIAGNOSTIC (`scenarios.py:1749`); injection hook present, not run/asserted in CI-short. Deferred by design.
- [AC-20] S10 blocking `claims_validator` — DIAGNOSTIC (`scenarios.py:1760`); **no setup hook and no blocking validator is wired into the RS**, so even when run it would not exercise the event-loop stall it names. Deferred by design (weakest scaffold).
- [AC-21] S11 memory/FD soak — NIGHTLY (`scenarios.py:1770`); no RSS/FD measurement implemented (design §5 metric absent). Deferred by design.
- [AC-22] S12 multi-tenant + issuer mix-up — NIGHTLY (`scenarios.py:1780`); drives `valid`+`wrong_iss` but no `dct`/`tenants` LRU-survival or RFC 9207 assertion. Deferred by design.
- [AC-24] SLO gates + baseline calibration — `Gate`/`GATES` machinery exists (`runner.py:1239/1255`) but **all thresholds are `None`** (permissive); only zero-500 + steady-state-divergence are enforced. Numeric SLO bars (design §5) and baseline calibration are deferred to the test/docs phases.

### SCOPE CREEP
- None material. `cache_metrics.py`, the `/metrics` route (`rs_app.py:93`), and mock-OP `RequestStats` (`mock_op.py:736`) all trace to design §2/§5 ("Cache instrumentation … TH-1.5 needs it" / "upstream fetches/issuer"). The T299 cache-counter work is a design-mandated dependency of TH-1.5, not out-of-scope.

### Architecture Violations
- [§5 metrics] `p999` latency, RSS/FD-over-time, and event-loop lag are not reported. RSS/FD/lag are nightly-tier (acceptable deferral); **p999 is a cheap omission absent even from the CI-short summary** despite being an explicit §5 metric (`locustfile.py:92-94` stops at p99).
- [§4 tiering — doc inconsistency] Design §4 states CI-short = S1–S8; the implementation moves S5→DIAGNOSTIC and S7→NIGHTLY (intentional, per task note), but `scenarios.py:164` and `test_load_ci_short.py:6` docstrings still say "S1-S8"/"CI_SHORT (S1-S8)", internally contradicting the `Profile` assignments (actual CI-short = S1,S2,S3,S4,S6,S8). Cosmetic; behaviour is data-driven and correct.
- [`locustfile.py:14` docstring] Claims the summary carries "per-class … latency percentiles"; it carries per-class counts only (drives AC-12 gap above).

### Summary
- Total ACs: 25
- Pass: 13 | Fail: 0 | Partial: 12

jamescrowley321:
### Review agent: review-sentinel

None.

### WARN (should fix)
None.

### INFO (acceptable risk)
- `src/py_identity_model/aio/token_validation.py:138-164, 237-261, 300-309` — The T299 counter hooks are purely additive. Each `get_cache_counters().record_*()` call is inserted alongside existing logic and returns nothing that is branched on. The `if response.is_successful:` guards wrap ONLY the counter increment; `apply_disco_cache_outcome` / `apply_jwks_cache_outcome` and the function return value still execute unconditionally afterward. Verified against the shipped file (not just the diff): validation control flow, cache-write behavior, and return values are unchanged, so the counters cannot introduce a fail-open path. A `record_*` is an O(1) int increment under a leaf `threading.Lock`; if it ever raised it would propagate (fail-closed, no token accepted), never fail-open. Fail-closed coverage exists in `src/tests/security/test_aio_cache_fetch_controls.py` (scheme-rejection asserts counters stay 0 AND `route.call_count == 0`, cooldown-sidecar eviction, coalescing guard, double-checked-lock recency) — deleting the scheme/policy control breaks those tests.

- `src/tests/harness/rs_app.py:93-90 (/metrics route)` — Unauthenticated `/metrics` explicitly appended to the middleware `excluded_paths`. Acceptable: (1) it exists ONLY in the test-harness RS app — grep confirms no `/metrics` route in `packages/fastapi-identity-model/` or the shipped library; the library exposes only the `get_cache_counters()` read API, no HTTP surface. (2) The payload is `get_cache_counters().snapshot()`, five aggregate per-process integers (`disco_hits/misses`, `jwks_hits/misses/refreshes`). No tokens, issuers, URIs, kids, or PII. The counters are process-wide aggregates, not per-request, so they are not a token-validity oracle. The exclusion is narrowly scoped to the literal `/metrics` path and does not shadow `/protected`.

- `src/tests/harness/mock_op.py:826-833 (/_stats, /_stats POST reset), _handle_control` — Unauthenticated stats read/reset and failure-injection control endpoints on the mock OP. Acceptable: the mock OP is a test double, never deployed, and `/_stats` returns only per-path request counts.

- `tools/mutation_security_allowlist.txt:2704-2741` — 20 mutants allowlisted for `aio.token_validation`. Reviewed: all are class-1 (log-message text of the "Forcing JWKS refresh" / empty-keys-fallback info lines — no control flow, return, or counter effect) or class-2 (`hash(None) % 32` lock-stripe selection — same-URI fetches still serialize on one lock object; only which of 32 stripes is chosen changes, unobservable). None mask a real security control; the SSRF/scheme, cooldown-eviction, and coalescing controls remain gated and are covered by fail-closed tests. Justified.

### Summary
- BLOCK: 0 | WARN: 0 | INFO: 4
- Overall: PASS

jamescrowley321:
### Review agent: review-viper

## Review: Red Team (Viper)

### Attack Surface
- **`src/py_identity_model/core/cache_metrics.py`** (NEW, SHIPPED) — process-global `CacheCounters` singleton, exported at top level via `get_cache_counters()`. The only genuinely shipped new code. Exposes aggregate integer hit/miss/refresh counters. No token, key, tenant, or claim data is stored.
- **`src/py_identity_model/aio/token_validation.py`** (SHIPPED) — added `record_*` counter calls to `_get_disco_response` / `_get_cached_jwks` / `_refresh_jwks`. Purely additive instrumentation; no change to signature verification, issuer/audience checks, scheme enforcement, or cache-write control flow.
- **`src/tests/harness/rs_app.py`** (TEST-ONLY) — new unauthenticated `GET /metrics` route added to the middleware `excluded_paths`.
- **`src/tests/harness/mock_op.py`** (TEST-ONLY) — new unauthenticated `GET/POST /_stats` endpoints alongside the pre-existing unauthenticated `POST /_control` failure-injection endpoint. Test double.
- **`src/tests/load/*`** (TEST-ONLY) — Locust replay driver; `runner.py` spawns `python -m locust` via `subprocess.run` with a fixed argv (no shell).
- **`uv.lock` / `pyproject.toml`** — `locust` (+ gevent, greenlet, flask, werkzeug, blinker, brotli, bidict, zope-*, configargparse, flask-cors, flask-login) added under `[dependency-groups].load`.

### Findings

No exploitable findings.

### Non-exploitable observations (for completeness, not findings)

- **New dependencies never reach shipped code.** `locust` and its transitive tree are declared in `[dependency-groups].load`, which is a PEP 735 dev dependency-group — it is NOT `[project.optional-dependencies]` and NOT a runtime dependency, so it is never resolved for consumers who `pip install py-identity-model`. The `uv_build` backend packages only `src/py_identity_model/`; `src/tests/` (harness, load, mock OP, rs_app) is not in the wheel/sdist. Locked versions are current and patched (locust 2.46.3, werkzeug 3.1.8 — past the 3.0.6 multipart-DoS/path CVEs, gevent 26.7.0, greenlet 3.5.5, flask 3.1.3). No known unpatched CVE, and even a hypothetical one could not reach a production install. Verified no module under `src/py_identity_model/` imports the harness/load packages.
- **Unauthenticated `GET /metrics` on rs_app is test-only and leaks nothing sensitive.** `rs_app.py` lives under `src/tests/harness/`, is not packaged, and is booted only by the load runner bound to a local uvicorn subprocess. The readout is `get_cache_counters().snapshot()` — five aggregate integers (disco/jwks hits/misses/refreshes). No token, key material, subject, tenant, or claim is exposed, so even if a downstream consumer copied this pattern into production the disclosure is limited to cache-efficiency counters (informational). It is O(1), so no DoS lever.
- **Mock OP `/_control` and `/_stats` are unauthenticated but test-only.** The mock OP is a deliberately controllable test double served in-process/localhost during the harness run; its failure-injection and stats routes are never shipped and never bound to a routable interface. Not reachable by an external attacker.
- **The `if response.is_successful` counter gate is not a security control.** It only prevents scheme-rejected (non-HTTPS) or coalesced fetches from inflating the miss/refresh tally. The cache-write (`apply_*_cache_outcome`) and all validation logic are unchanged, so a wrong gate would at worst skew a metric, not affect auth. HTTPS scheme enforcement continues to run inside `get_discovery_document`/`get_jwks` exactly as before; new tests in `src/tests/security/test_aio_cache_fetch_controls.py` pin that behaviour against mutation.

### Summary
- Attack surface elements: 6 (1 shipped counter module, 1 shipped instrumentation change, 4 test-only/dev-only)
- Overall: PASS

jamescrowley321:
### Review agent: review-delta-blind

## Review: Blind Hunter

Delta re-review of the fix commit. Verified each addressed finding and scanned the fix diff for new regressions.

### MUST FIX
No findings.

### SHOULD FIX
No findings.

### NITPICK
- [`src/tests/load/runner.py:288`] `alg_cost_ratio` treats a `p95` of `0.0` as "absent" (`if not num or not den`) and returns `None`; a legitimately sub-millisecond warm class would make `test_s2_alg_cost_ratio_is_reportable`'s `assert ratio is not None` fail. Harmless in practice (crypto-validation p95 is never 0ms), so noted only.
- [`src/tests/load/locustfile.py:84`] inner `by_class` dict is annotated `dict[str, float]` but stores `entry.num_requests`/`num_failures` (ints) alongside the float percentiles — cosmetic type-hint drift, no runtime effect.

---

### Verification notes (fixes confirmed correct)

1. **miss/refresh gating on `_upstream_fetch_attempted`** — CONFIRMED correct.
   - `ConfigurationException` is imported (`aio/token_validation.py:49`) and `validate_url_scheme` (`aio/token_validation.py:15`) is the exact same pure pre-flight `get_discovery_document`/`get_jwks` run before any network I/O (`aio/discovery.py:54`, `aio/jwks.py:59`), so the re-parse faithfully mirrors the fetch's short-circuit condition.
   - Same `policy`/`address` objects are passed to both the fetch and `_upstream_fetch_attempted` in all three call sites (`_get_disco_response:184-195`, `_get_cached_jwks`, `_refresh_jwks`) — no divergence.
   - `get_discovery_document`/`get_jwks` catch **all** exceptions (`except Exception: return handle_*_error(e)`), returning an unsuccessful response on 429/5xx/connection error, so the `if _upstream_fetch_attempted(...)` line is reached and the failed round trip is counted — matching the new docstring intent.
   - Existing fail-closed security tests (`test_aio_cache_fetch_controls.py:110-161`) that assert scheme-rejected http:// requests keep misses/refreshes at 0 remain satisfied (scheme check raises → returns `False` → no increment). No test contradiction introduced.
   - Retries collapse to one logical fetch (single `get_*` call), so misses are counted once, as the new tests assert (`test_cache_metrics.py:248,262`).

2. **`cache_hit_rate` returns `0.0` on `total == 0`** — CONFIRMED (`runner.py:317`). Coherent with fix #1: an all-error window now records failed fetches as misses so `total != 0` there; the `total == 0` guard only fires when nothing reached the counted paths, and 0.0 (not 1.0) correctly prevents an outage/broken scrape from masquerading as a warm cache. "cold" gate (S4/S3) has all thresholds unset, so no false gate failure.

3. **rs_app imports `_default_excluded_paths`** — CONFIRMED. Function exists (`config.py:9`) and returns a fresh list, so the subsequent `.append(_METRICS_PATH)` mutates a copy, not the package default. Import path matches module layout.

4. **per-class latency + p999** — CONFIRMED. `p999` plumbed through summary → `_to_result:325`; `latency_by_class` built from present p50/p95/p99 keys; S2 scenario carries both `valid` and `valid_es256` classes (`scenarios.py:196`) that the new test exercises.

5. **S4 → NIGHTLY, S10 scaffold** — CONFIRMED. `"cold"` is a valid gate key (`runner.py:73`); `_short_ttl` sets `discovery_cache_control`/`jwks_cache_control`, both real `MockOPControls` fields (`mock_op.py:143-144`). Scenario prose/CI-short catalogue and its assertion (`test_every_ci_short_scenario_ran`) updated consistently.

jamescrowley321:
### Review agent: review-delta-edge-case

## Review: Edge Case Hunter

### Findings

No unhandled edge cases found.

Verification notes (all fix paths traced, no new regressions):

- `_upstream_fetch_attempted` (aio/token_validation.py:42-46): re-parses the
  **same** `DiscoveryPolicy` the fetch used. `ConfigurationException` is imported
  (line 49) and is the only exception `validate_url_scheme` raises; it is
  side-effect-free (pure `urlparse`), so the re-check yields the identical verdict
  as the pre-flight the fetch applied. Empty/None address → both fetch and
  re-check reject consistently (no miscount). 429/5xx/connection-error → scheme
  check passes → miss counted, matching intent.
- All three call sites (`_get_disco_response` :184, `_get_cached_jwks` :285,
  `_refresh_jwks` :332) construct `policy = DiscoveryPolicy(require_https=...)`
  locally and pass the same object to both the request and the accounting call —
  policy is never None, so no `AttributeError` path.
- `_refresh_jwks` coalesced early return (:328-329) returns before reaching the
  counter; only the GET-issuing branch increments. Correct.
- `cache_hit_rate` total==0 → 0.0 (runner.py:317): all `GATES` thresholds are
  `None` (permissive, runner.py:71-74) so `min_cache_hit_rate` gating is inert;
  no scenario assertion regresses. total==0 only occurs when nothing reached the
  counted cache paths (failed fetches now count as misses), so 0.0 is the correct
  signal.
- `alg_cost_ratio` (runner.py:288-292): guards `not num or not den`, covering
  both None (absent class) and 0.0 p95 (degenerate sample). No div-by-zero.
- Per-class percentiles / `p999`: `summary.get(...)` and `by_class[name][p]`
  reads are same-run producer/consumer (locustfile writes p50/p95/p99); `p999`
  guarded with default 0.0. No key-miss crash.
- rs_app.py `_default_excluded_paths()` returns a **fresh** list per call
  (config.py:9-10), so `excluded.append(_METRICS_PATH)` mutates a local list, not
  shared module state. No aliasing bug.

### Summary
- Unhandled paths found: 0

jamescrowley321:
### Review agent: review-delta-acceptance

## Review: Acceptance Auditor

Delta re-review of the TH-1.5 FIX commit. Scope = the five fix claims (AC-a..AC-e). Each claim verified against `review-delta.patch` plus the live source/tests under `/tmp/pim-T311/`.

### PASS
- [AC-a] Failed upstream fetches (429/5xx/conn error) count as a cache miss/refresh — implemented at `src/py_identity_model/aio/token_validation.py:56-84` (`_upstream_fetch_attempted`, gating on the pre-flight scheme check instead of `response.is_successful`) and wired at `token_validation.py:188-196`, `:284-292`, `:331-339`. Tested at `src/tests/unit/test_cache_metrics.py:239` (503 disco → `disco_misses == 1`), `:254` (429 jwks → `jwks_misses == 1`), `:268` (502 refresh → `jwks_refreshes == 1`) (tier: unit). Caches + counters reset per test via `_clean_cache_state` (`test_cache_metrics.py:158`), so each asserts a real fetch. Reverting to `if response.is_successful` makes all three counters read 0 → all three tests fail. Confirmed removal-sensitive.
- [AC-b] S2 RS256-vs-ES256 alg-cost ratio computable from per-class latency — implemented at `src/tests/load/runner.py:99-110` (`LoadResult.alg_cost_ratio`, p95 ratio, returns `None` on missing/zero) fed by per-class p50/p95/p99 emitted at `src/tests/load/locustfile.py:86-93` and mapped into `latency_by_class` at `runner.py:257-260`. S2 carries both classes (`classes=("valid", "valid_es256")`, `scenarios.py:194-199`). Tested at `src/tests/load/test_load_ci_short.py:57` (`test_s2_alg_cost_ratio_is_reportable`) — asserts both classes drove load and `alg_cost_ratio(...)` is a positive finite number (tier: integration, real Locust subprocess run). If `latency_by_class` were dropped, `alg_cost_ratio` returns `None` and the test fails. Confirmed.
- [AC-d] CI_SHORT honestly = S1,S2,S3,S6,S8; S4→NIGHTLY; S10 scaffold — profile membership in `src/tests/load/scenarios.py` (S1 :187, S2 :195, S3 :203, S6 :238, S8 :259 = CI_SHORT; S4 :212 now `Profile.NIGHTLY`; S10 :280 DIAGNOSTIC). S4 justification is truthful: `MIN_CACHE_TTL_SECONDS = 60.0` with clamp at `core/jwks_cache.py:41,440,456` — a 60s floor genuinely cannot roll over in a ~3s window. S10 docstring is honestly downgraded ("INCOMPLETE… placeholder… Do NOT treat as coverage", `scenarios.py:284-289`) — no overstated coverage. Exercised by `test_every_ci_short_scenario_ran` and the parametrized gate run (`test_load_ci_short.py:52,72`, tier: integration). Note: `_CI_SHORT_IDS` is derived from `profile_scenarios(CI_SHORT)` so that assertion is somewhat tautological, but the concrete `profile=` data is what fixes membership and it is correct + truthful.

### FAIL
- (none)

### PARTIAL
- [AC-c] p999 metric now captured — DONE: emitted at `src/tests/load/locustfile.py:102` (`get_response_time_percentile(0.999)`), carried through `LoadResult.p999_ms` (`runner.py:89`) and populated at `runner.py:249`. MISSING: no test asserts on p999. `p999_ms` construction uses `summary.get("p999", 0.0)`, so removing the locustfile emission silently yields `0.0` and breaks no test — the metric is captured but not removal-verified. "Captured" is truthful; "tested" is not.
- [AC-e] `cache_hit_rate` no longer masks an empty window as 100% — DONE: `runner.py:132` returns `hits / total if total else 0.0` (was `else 1.0`). MISSING: no test exercises the `total == 0` branch. The only `cache_hit_rate` assertion is `s1.cache_hit_rate >= 0.99` (`test_load_ci_short.py:112`), which is a warm window (`total > 0`) and passes identically under `else 1.0` or `else 0.0`. No CI_SHORT gate sets `min_cache_hit_rate` on an empty window. Reverting the fix to `else 1.0` breaks no test. Code fix is real and correct; the anti-masquerade behavior is untested.

### SCOPE CREEP
- `src/tests/harness/rs_app.py:28,74` — replaces the hardcoded `_DEFAULT_EXCLUDED_PATHS` list with a call to `fastapi_identity_model.config._default_excluded_paths()`. Not one of the five fix claims (AC-a..AC-e); a drift-hardening refactor. Benign and defensible, but not traceable to a stated AC and carries no test of its own.

### Architecture Violations
- (none) — gating on the pure, side-effect-free `validate_url_scheme` re-parse keeps the counter decision in the sync/async wrapper without reaching into core I/O; forged non-https URIs still excluded, matching the documented invariant in `core/cache_metrics.py` docstring (which was updated to match, not overstated).

### Summary
- Total ACs: 5 | Pass: 3 | Fail: 0 | Partial: 2

Notes on truthfulness: AC-a/AC-b/AC-d claims hold and are removal-verified/honest. AC-c and AC-e are genuinely implemented in code but the fix claims imply verification the tests do not provide — reverting either leaves the whole suite green. No false docstring found; the `cache_metrics.py` and `scenarios.py` S10 docstrings were correctly revised to match reality.

